Issue #22, opened 2020-03-30
Update dependencies


jowerner:
Several XLT dependencies are available in a newer version.

jowerner:
Updated most of the outdated compile/runtime dependencies, except:

* tyrus 1.16 -> pulls dependencies that are more dated than before
* dnsjava 3.x -> may require code changes (see #24)

Exclusion rules have been updated as well.

h-arlt:
Found some updates for the following dependencies:
* aws-sdk-java-ec2 1.11.762
* xstream 1.4.12

Test deps:
* archunit-junit4 0.13.1
* pmd-java 5.8.1 (latest is 6.22.0 which is potentially incompatible)
* checkstyle 6.19 (latest is 8.31 which is potentially incompatible)

The used mocking frameworks also have some new versions but updating them should be done very carefully. This might be part of a separate issue.

jowerner:
Done. I also adjusted the configuration of Checkstyle and PMD to get them working again.

The mocking libs we will update in #28 as the transition to Mockito 2.x requires some code changes.

h-arlt:
I noticed that there are several `asm*` artifacts that don't have the same version. It seems that it does not cause any problems right now but I don't have a good feeling as I did not test all functionalities thoroughly.

I was able to unify all `asm*` artifacts to version 7.1 by removing `asm-commons` (not needed at all) and updating `parboiled-java` (dependency of pegdown). 
See PR #29.

jowerner:
Our latest `asm` updates break the Cobertura coverage report ...

PR #29, opened 2020-04-17
Unify asm versions

h-arlt:
As of issue #22, there are several `asm*` artifacts that don't have the same version. This might lead to compatibility issues as there are dependencies between those artifacts.


jowerner:
Works for now. If issues arise later on (for instance with PMD), we might pin the asm* libs to a certain version by defining them as top-level dependencies.

h-arlt:
Agreed.

Issue #54, opened 2020-09-01
Non-interactive mode for the master controller


jowerner:
In order to better fit into highly-automated environments, the master controller should be improved for better scriptability. Currently, the master controller can be used either in interactive mode or in auto mode. None of them is suited when tasks like the following should be carried out in a scripted manner:

* upload the test suite and start the load test
* download the results of a running load test and generate a report from them
* abort a running load test
* abort a running load test, download the final results, and generate a report from them

What we need is a non-interactive mode where the user can configure on the command line what is to be done.

jowerner:
See branch "MasterControllerCommands".

jowerner:
Branch merged.

PR #55, opened 2020-09-03
#54 Non-interactive mode for MasterController

h-arlt:


src/main/java/com/xceptance/xlt/mastercontroller/Main.java
```diff
@@ -422,21 +443,75 @@ private static CommandLine parseCommandLine(final String[] args)
...
+        System.out.println("    mastercontroller [<other options>]");
+        System.out.println("      -> Runs in interactive mode. Choose the command to be executed next.");
         System.out.println();
-        formatter.printHelp(XltConstants.MASTERCONTROLLER_EXECUTABLE_NAME + " [<options>]", "\nOptions:", options, "");
+        System.out.println("    mastercontroller -c <commandList> [<other options>]");
+        System.out.println("      -> Runs in non-interactive mode. Pass the commands to be executed.");
+        System.out.println();
+        System.out.println("    mastercontroller -auto [<other options>]");
+        System.out.println("      -> Runs a load test in non-interactive mode by executing all needed commands automatically.");
+        System.out.println();
```
h-arlt, line 456: Why not using the formatter for printing that as was done before? It's a little bit weird to format and print the program usage text manually but use the formatter to format and print the program's options.

src/main/java/com/xceptance/xlt/mastercontroller/Main.java
```diff
@@ -422,21 +443,75 @@ private static CommandLine parseCommandLine(final String[] args)
...
+     */
+    private static void validateCommandLine(final CommandLine commandLine)
+    {
+        boolean invalid = false;
+
+        // -auto|-faf|-sequential and -c are mutually exclusive
+        if ((commandLine.hasOption(OPTION_AUTO) || commandLine.hasOption(OPTION_FAF) || commandLine.hasOption(OPTION_SEQUENTIAL)) &&
+            commandLine.hasOption(OPTION_COMMANDS))
+        {
+            System.out.printf("Option '-%s' cannot be used together with '-%s', '-%s', or '-%s'.\n", OPTION_COMMANDS, OPTION_AUTO,
```
h-arlt, line 483: Log at level ERROR additionally?

src/main/java/com/xceptance/xlt/mastercontroller/Main.java
```diff
@@ -422,21 +443,75 @@ private static CommandLine parseCommandLine(final String[] args)
...
+
+        // check -c option value
+        final String commandList = commandLine.getOptionValue(OPTION_COMMANDS);
+        if (commandList != null)
+        {
+            final String[] unknownCommands = MasterControllerCommands.validate(commandList);
+            if (unknownCommands.length > 0)
+            {
+                System.out.printf("Unknown commands given with the '-%s' option: %s\n", OPTION_COMMANDS,
+                                  StringUtils.join(unknownCommands, ", "));
```
h-arlt, line 496: Log at level ERROR additionally?

h-arlt (commented):
Besides a few small improvements/suggestions overall code looks good to me.

Q: Do we have to point out (in the help text, for example) that the passed commands are executed in the same order as given on the command line (even if they don't make sense)?

src/main/java/com/xceptance/xlt/mastercontroller/Main.java
jowerner: True. Fixed.

jowerner:
> Q: Do we have to point out (in the help text, for example) that the passed commands are executed in the same order as given on the command line (even if they don't make sense)?

Yep. Added this to the help text.

src/main/java/com/xceptance/xlt/mastercontroller/Main.java
jowerner: Done.

h-arlt:
Merge at will ;-)

PR #64, opened 2020-10-29
separated status calculation from status display for better maintaina…

jowerner:
…bility and extensibility


h-arlt (approved):
As the received agent status set is looped over multiple times now, we might need to keep an eye on its performance, esp. when running load tests with hundreds of agents.

PR #65, opened 2020-10-29
Allow for subclassing of master controller Main

jowerner:
- made methods of master controller Main non-static and protected so we can inherit from Main and overwrite these methods as needed
- moved any System.exit() to the new run() method so the JVM does not quit by accident when subclasses call any of the other methods
- adjusted test case

### DIFF
--- a/src/main/java/com/xceptance/xlt/mastercontroller/Main.java
+++ b/src/main/java/com/xceptance/xlt/mastercontroller/Main.java
@@ -75,7 +75,7 @@ public class Main
 
     private static final Log log = LogFactory.getLog(Main.class);
 
-    private BasicConsoleUI ui;
+    protected BasicConsoleUI ui;
 
     protected void initialize(final CommandLine commandLine) throws Exception
     {
@@ -163,7 +163,7 @@ public class Main
      * @param config
      *            master controller configuration
      */
-    private void setupHttpsProxy(final MasterControllerConfiguration config)
+    protected void setupHttpsProxy(final MasterControllerConfiguration config)
     {
         if (config.isHttpsProxyEnabled())
         {
@@ -183,7 +183,7 @@ public class Main
      *            master controller configuration
      * @return the HessianProxyFactory
      */
-    private HessianProxyFactory getHessianProxyFactory(final MasterControllerConfiguration config)
+    protected HessianProxyFactory getHessianProxyFactory(final MasterControllerConfiguration config)
     {
         final HessianProxyFactory proxyFactory = new EasyHessianProxyFactory();
         proxyFactory.setConnectTimeout(config.getAgentControllerConnectTimeout());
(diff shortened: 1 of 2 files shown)


src/main/java/com/xceptance/xlt/mastercontroller/Main.java
```diff
@@ -163,7 +163,7 @@ private static File getOverridePropertieFile(final CommandLine commandLine)
      * @param config
      *            master controller configuration
      */
-    private static void setupHttpsProxy(final MasterControllerConfiguration config)
+    private void setupHttpsProxy(final MasterControllerConfiguration config)
```
h-arlt, line 166: Shouldn't this be `protected` to allow overrides?

src/main/java/com/xceptance/xlt/mastercontroller/Main.java
```diff
@@ -183,7 +183,7 @@ private static void setupHttpsProxy(final MasterControllerConfiguration config)
      *            master controller configuration
      * @return the HessianProxyFactory
      */
-    private static HessianProxyFactory getHessianProxyFactory(final MasterControllerConfiguration config)
+    private HessianProxyFactory getHessianProxyFactory(final MasterControllerConfiguration config)
```
h-arlt, line 186: Shouldn't this be protected to allow overrides?

src/main/java/com/xceptance/xlt/mastercontroller/Main.java
```diff
@@ -201,7 +201,7 @@ private static HessianProxyFactory getHessianProxyFactory(final MasterController
      *            master controller configuration
      * @return the UrlConnectionFactory
      */
-    private static UrlConnectionFactory getUrlConnectionFactory(final MasterControllerConfiguration config)
+    private UrlConnectionFactory getUrlConnectionFactory(final MasterControllerConfiguration config)
```
h-arlt, line 204: Shouldn't this be protected to allow overrides?

src/main/java/com/xceptance/xlt/mastercontroller/Main.java
```diff
@@ -75,12 +75,12 @@
 
     private static final Log log = LogFactory.getLog(Main.class);
 
-    private static BasicConsoleUI ui;
+    private BasicConsoleUI ui;
```
h-arlt, line 78: Allow subclasses to access this field (w/o the need to override `setupUi`)?

h-arlt (approved):
Some methods - and the `ui` field - might also be declared `protected` but not necessarily.

src/main/java/com/xceptance/xlt/mastercontroller/Main.java
jowerner: Done.

src/main/java/com/xceptance/xlt/mastercontroller/Main.java
jowerner: Done.

src/main/java/com/xceptance/xlt/mastercontroller/Main.java
jowerner: Done.

src/main/java/com/xceptance/xlt/mastercontroller/Main.java
jowerner: Done.

PR #75, opened 2021-02-10
74 enable http request for file uploads

jowerner:


src/main/java/com/xceptance/xlt/engine/httprequest/HttpRequest.java
```diff
@@ -362,22 +362,41 @@ public HttpRequest removeHeaders()
      */
     public HttpRequest param(final String name, String value)
     {
+        param(new NameValuePair(name, value));
+
+        return this;
```
h-arlt, line 367: As this is just a delegate to the overloaded `param` method, I'd prefer to change its implementation to
`return param(new NameValuePair(name, value));`
WDYT?

src/main/java/com/xceptance/xlt/engine/httprequest/HttpRequest.java
```diff
@@ -362,22 +362,41 @@ public HttpRequest removeHeaders()
...
+     * @param nameValuePair
+     *            the name/value pair representing the parameter to add
+     */
+    public HttpRequest param(NameValuePair nameValuePair)
+    {
+        String name = nameValuePair.getName();
+
         if (StringUtils.isBlank(name))
         {
             throw new IllegalArgumentException("Parameter name must not be blank.");
```
h-arlt, line 382: Message is a little bit confusing as there is no 'name' parameter. Hence, I'd change to something like "NameValuePair parameter is invalid since its name is blank".

src/main/java/com/xceptance/xlt/engine/httprequest/HttpRequest.java
```diff
@@ -390,7 +409,7 @@ public HttpRequest param(final String name, String value)
      */
     public HttpRequest params(final List<NameValuePair> params)
     {
-        params.forEach(p -> param(p.getName(), p.getValue()));
+        params.forEach(p -> param(p));
```
h-arlt, line 412: Might be shortened by using method reference: `params.forEach(this::param);`
WDYT?

h-arlt (commented):
Overall code looks good. I added some potential improvements that can be applied but don't have to.

src/main/java/com/xceptance/xlt/engine/httprequest/HttpRequest.java
jowerner: True.

src/main/java/com/xceptance/xlt/engine/httprequest/HttpRequest.java
jowerner: Done.

src/main/java/com/xceptance/xlt/engine/httprequest/HttpRequest.java
jowerner: Hm, difficult. The message must fit for all overloads, also for those that do not deal with NameValuePair directly. You know? And if you read the message as "Parametername ..." instead of "Parameter 'name' ...", the message will sound much better, eh? ;-)

h-arlt: Tricky, indeed. What about to switch the first two words: "Name of parameter must not be blank"?

jowerner: Done. Rephrased the other message in this method accordingly.

PR #76, opened 2021-02-10
33 binary put post support in http request

jowerner:


src/main/java/com/xceptance/xlt/engine/httprequest/HttpRequest.java
```diff
@@ -445,14 +457,71 @@ public HttpRequest removeParams()
...
+     *            the file from which to read the request body
+     * @throws IOException
+     *             if the file could not be read
+     */
+    public HttpRequest body(final File file) throws IOException
+    {
+        final byte[] bytes = FileUtils.readFileToByteArray(file);
+        body(bytes);
+
+        return this;
```
h-arlt, line 506: Can be shortened to `return body(bytes);` but doesn't have to.

src/main/java/com/xceptance/xlt/engine/httprequest/HttpRequest.java
```diff
@@ -445,14 +457,71 @@ public HttpRequest removeParams()
...
+     *            the input stream from which to read the request body
+     * @throws IOException
+     *             if the input stream could not be read
+     */
+    public HttpRequest body(final InputStream inputStream) throws IOException
+    {
+        final byte[] bytes = IOUtils.toByteArray(inputStream);
+        body(bytes);
 
         return this;
```
h-arlt, line 526: Can be shortened to `return body(bytes);` but doesn't have to.

h-arlt (approved):
Well done.

src/main/java/com/xceptance/xlt/engine/httprequest/HttpRequest.java
jowerner: Done.

src/main/java/com/xceptance/xlt/engine/httprequest/HttpRequest.java
jowerner: Done.

Issue #86, opened 2021-04-27
Use separate logger for internal DomUtils class


h-arlt:
In order to be able to dump the parsed HTML documents as part of the XLT Result Browser, XLT creates clones of the parsed documents with the help of an internal class, namely `com.xceptance.xlt.engine.resultbrowser.DomUtils`. 
This class logs a message at level WARN to XLT's runtime logger each time a DOM operation fails, e.g. when setting of an attribute failed. This can easily blow up the log file with superfluous messages and there is no way for users to get rid of them (unless setting the log level of XLT's runtime logger to ERROR or FATAL).

To overcome this issue, the class `DomUtils` should its own logger whose level is set to WARN per default. Users can then configure this logger separately, e.g. set level to DEBUG or even OFF.

PR #87, opened 2021-04-27
#86: Use separate logger for internal DomUtils class

h-arlt:


src/main/java/com/xceptance/xlt/engine/resultbrowser/DomUtils.java
```diff
@@ -47,6 +47,11 @@
  */
 final class DomUtils
 {
+    /**
+     * Class logger.
+     */
+    private static final Logger LOGGER = Logger.getLogger(DomUtils.class);
```
jowerner, line 53: When creating class-level loggers elsewhere, we usually use the commons-logging approach:

```private static final Log LOG = LogFactory.getLog(DomUtils.class);```

I am aware that the previously used runtime logger is also a plain log4j logger. However, this has been wrong all the time, but could not be changed because it's API.

WDYT about changing this to commons-logging style?

jowerner (changes requested):

src/main/java/com/xceptance/xlt/engine/resultbrowser/DomUtils.java
h-arlt: Changed to commons-logging.

jowerner (approved):

Issue #93, opened 2021-05-19
XLT keeps a hold on the properties files of a test suite


jowerner:
Once the Master Controller has uploaded a test suite to the Agent Controllers, it is no longer possible to edit the configuration files of the test suite, for example `project.properties`, on the local machine. The editor complains that the file is in use by another process.

A similar effect can be observed on the Agent Controller machine. Once the Master controller has downloaded the results of a test run, uploading a test suite with a modified properties file will fail on the Agent Controller, again with the error that the old properties file is in use by someone else on the Agent Controller machine.

Note that both Master Controller and Agent Controller load and inspect the properties files of a test suite when uploading the suite and when downloading results. Looks like XLT keeps a hidden handle to the properties files of a test suite. This behavior can be observed on Windows, but not on Linux.

jowerner:
When reading the properties files of a test suite, XLT did not close the properties files correctly.

PR #94, opened 2021-05-19
#93: XLT keeps a hold on the properties files of a test suite

jowerner:


src/main/java/com/xceptance/xlt/engine/util/IncludedFilesResolver.java
```diff
@@ -306,10 +306,9 @@ private static void checkForCyclicInclude(final List<String> alreadyMarked, fina
      */
     private static Properties getPropertiesFromFile(final FileObject current) throws IllegalStateException
     {
```
h-arlt, line 308: What about to simply delegate to `PropertiesUtils.loadProperties(FileObject)` inside the try-block?

jowerner: Good idea! Done.

Issue #100, opened 2021-06-02
NPE when listing or deleting machine instances with gce_admin


jowerner:
`gce_admin` is used to create, list, and delete machine instances in Google Compute Engine. When listing machine instances, the following error may be printed to the console:

```
Querying all instances in region 'europe-west3' ... OK

--- Master controller configuration ---
Failed to execute operation: null
```

When deleting instances, a similar error may occur: 

```
You selected to terminate *all* managed instances of group 'jw'

  in region: europe-west3 ... Failed: null
```

jowerner:
Both errors were triggered by a machine instance that does not have a public IP, but a private IP only. Now the code falls back to a private IP in case no public IP is assigned to the machine.

PR #101, opened 2021-06-02
#100: NPE when listing or deleting machine instances with gce_admin

jowerner:
* fall back to a private IP if an instance does not have a public IP
* be more defensive in this area as all values returned by Google might be null


h-arlt (approved):
Looks good to me.

Issue #104, opened 2021-06-03
Discard IPv6 addresses received from DNS


jowerner:
When resolving a host name via DNS, we usually get back one or more IPv4 addresses and maybe also IPv6 addresses. Sometimes using IPv6 causes issues. Maybe it is not enabled on your local machine or not correctly set up. In some cloud environments, IPv6 is not available either.

In order to work around such issues, we should enable XLT to optionally discard IPv6 addresses received from DNS and use the IPv4 addresses only. Likewise, it may come in handy to discard IPv4 addresses only, for testing purposes.

jowerner:
Now there are two new properties to configure the wanted behavior:

```
xlt.dns.ignoreIPv4Addresses = false
xlt.dns.ignoreIPv6Addresses = false
```

PR #106, opened 2021-06-03
#104: Discard IPv6 addresses received from DNS

jowerner:


src/main/java/com/xceptance/xlt/engine/dns/XltDnsResolver.java
```diff
@@ -99,6 +117,8 @@ public XltDnsResolver()
         recordAddresses = props.getProperty(PROP_RECORD_ADDRESSES, false);
         shuffleAddresses = props.getProperty(PROP_SHUFFLE_ADDRESSES, false);
         pickOneAddressRandomly = props.getProperty(PROP_PICK_ONE_ADDRESS_RANDOMLY, false);
+        ignoreIPv4Addresses = props.getProperty(PROP_IGNORE_IPV4_ADDRESSES, false);
+        ignoreIPv6Addresses = props.getProperty(PROP_IGNORE_IPV6_ADDRESSES, false);
```
h-arlt, line 122: Do we want to error out in case user has configured to ignore both, ipv4 and ipv6? Just to give some aid in case of misconfiguration - it doesn't make any sense to ignore both.

src/main/java/com/xceptance/xlt/engine/dns/XltDnsResolver.java
```diff
@@ -54,6 +58,10 @@
 
     private static final String PROP_RECORD_ADDRESSES = PROP_PREFIX_DNS + "recordAddresses";
 
+    private static final String PROP_IGNORE_IPV4_ADDRESSES = PROP_PREFIX_DNS + "ignoreIPv4Addresses";
+
+    private static final String PROP_IGNORE_IPV6_ADDRESSES = PROP_PREFIX_DNS + "ignoreIPv6Addresses";
+
```
h-arlt, line 64: What about to use `xlt.dns.ipv4` and `xlt.dns.ipv6` as property names? Having it configured to `false` would then simply mean that this protocol version is disabled, and `true` having it enabled respectively.

samples/testsuite-xlt/config/default.properties
```diff
@@ -109,12 +109,16 @@ com.xceptance.xlt.http.responseId.headerName = X-XLT-ResponseId
 ##                              whole lifetime of a WebClient
 ## - shuffleAddresses ......... whether to shuffle multiple resolved addresses
 ## - pickOneAddressRandomly ... whether to use only one out of multiple addresses
+## - ignoreIPv4Addresses ...... whether to ignore IPv4 addresses received from DNS
+## - ignoreIPv6Addresses ...... whether to ignore IPv6 addresses received from DNS
```
h-arlt, line 113: IMO there is no need/benefit in saying that these addresses where received from DNS since the entire config section is about DNS.

h-arlt (commented):
Some (more or less meaningful) remarks.

src/main/java/com/xceptance/xlt/engine/dns/XltDnsResolver.java
jowerner: Good idea. Now a warning is printed to the logs. Hope that suffices?

h-arlt: That should do it. Thanks!

src/main/java/com/xceptance/xlt/engine/dns/XltDnsResolver.java
jowerner: I like it better the way it is now. With plain "ipv4" and "ipv6" it may not be clear what is meant. I could also be related to the way DNS is contacted, etc. You know what I mean?

samples/testsuite-xlt/config/default.properties
jowerner: I hope being a bit more verbose here helps understanding better what exactly is meant.

h-arlt: Yep.

src/main/java/com/xceptance/xlt/engine/dns/XltDnsResolver.java
h-arlt: Roger.

Issue #110, opened 2021-06-15
Cookie headers not recorded with Chrome


jowerner:
When using real browsers such as Chrome or Firefox to execute test scenarios, XLT tries to record the details of all requests made with the help of special browser extensions. This includes also the request and response headers.

However, with Chrome/Chromium not all headers are recorded. For example, `Cookie` and `Set-Cookie` headers are missing.

Luckily, those headers are excluded for performance reasons only. By setting an additional flag in the Chrome extension, we will get access to them again. See https://developer.chrome.com/docs/extensions/reference/webRequest/ for more information.

BTW, no issues with Firefox, so nothing to do for the Firefox extension.

PR #111, opened 2021-06-15
#110: Cookie headers not recorded with Chrome

jowerner:
added flag to also get "extra" request/response headers such as Cookie and Set-Cookie


h-arlt (commented):
The `onBeforeRequest` event also accepts `extraHeaders` as option. Do we want to add it to our handler as well?

https://developer.chrome.com/docs/extensions/reference/webRequest/#type-OnBeforeRequestOptions

jowerner:
Since we do not process headers in the `onBeforeRequest` handler, I think we don't need that.

h-arlt (approved):

Issue #129, opened 2021-08-24
Optionally exclude logs and/or result browsers when downloading test results in non-interactive mode


h-arlt:
When running XLT mastercontroller in non-interactive mode, download of test results will always include logs and result browsers. This might be fine in most cases but sometimes log files and even result browsers may not be needed and/or users simply want to improve download speed by restricting the amount of test results.

For this reasons, XLT mastercontroller should offer an appropriate command line option (or multiple ones) that can be used to restrict the download to certain data types and is effective in non-interactive mode only, e.g. `noLogs` and `noResultBrowsers`. Please notice that `noResultBrowsers` also implies `noLogs`.

rschwietzke:
I suggest to avoid including something indirectly in another option. So noResultBrowsers must not exclude logs, because the name does not imply that. The command line does not have to show the same logic as the UI does. Once implemented we can never change that again easily.

The `noDownload` option already says download nothing. I suggest to make it clearer what the options do, because `noLogs` does not indicate anything in regards to a download for instance. Suggestion: `noResultBrowserDownload` and `noLogsDownload` which fits `noDownload` but they are all not really nicely named options.

Please don't forget that there is a `commands` option that needs different options too.

h-arlt:
Thanks for your feedback @rschwietzke .

Then let's start with a single new option named `download` that takes a comma-separated list of data types as argument. The argument is not optional. Possible values are `logs`, `measurements` and `resultbrowsers`.
In case the option is not given, then all data types will be downloaded. This is equivalent to specifying all data types as option value. 
The options `download` and `noDownload` are mutually exclusive. XLT mastercontroller will exit with an error in case both were given.

PR #131, opened 2021-08-26
#129: Option to configure download result amount in non-interactive mode

h-arlt:
Implemented new option `--only-download` that can be used to restrict result download to the given data types (comma-separated list). Supported values are `logs`, `measurements` and `resultbrowsers`.

Example: `mastercontroller.sh -auto -embedded --only-download=measurements,logs`

### DIFF
--- a/src/main/java/com/xceptance/xlt/agentcontroller/TestResultAmount.java
+++ b/src/main/java/com/xceptance/xlt/agentcontroller/TestResultAmount.java
@@ -23,10 +23,10 @@ public enum TestResultAmount
     ALL("Measurements, result browser data, and logs [all]", "1"),
     MEASUREMENTS_AND_RESULTBROWSER("Measurements and result browser data", "2"),
     MEASUREMENTS_AND_LOGS("Measurements and logs", "3"),
-    MEASUREMENTS_ONLY("Measurements only", "3"),
-    RESULTBROWSER_AND_LOGS("Result browser data and logs", "4"),
-    RESULTBROWSER_ONLY("Result browser only", "5"),
-    LOGS_ONLY("Logs only", "6"),
+    MEASUREMENTS_ONLY("Measurements only", "4"),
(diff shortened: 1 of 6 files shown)


src/main/java/com/xceptance/xlt/agentcontroller/TestResultAmount.java
```diff
@@ -22,7 +22,11 @@
 {
     ALL("Measurements, result browser data, and logs [all]", "1"),
     MEASUREMENTS_AND_RESULTBROWSER("Measurements and result browser data", "2"),
+    MEASUREMENTS_AND_LOGS("Measurements and logs", "3"),
     MEASUREMENTS_ONLY("Measurements only", "3"),
```
jowerner, line 26: The key "3" is already used. Please ensure all keys are unique.

h-arlt: Fixed.

src/main/java/com/xceptance/xlt/mastercontroller/Main.java
```diff
@@ -490,6 +519,44 @@ protected boolean validateCommandLine(final CommandLine commandLine)
...
+            {
+                final String message = String.format("Option '--%s' can only be used in non-interactive mode.", OPTION_DOWNLOAD);
+                System.out.println(message);
+                log.error(message);
+
+                invalid = true;
+            }
+            else
+            {
+                // validate option values
```
jowerner, line 545: Maybe we need to check first if the option value has a value at all. 
Consider passing the option in the following variants:

```
--only-download=
--only-download ""
```

In these cases, the download currently fails with an exception:

```
Failed to run master controller: Downloading test results failed.
```

h-arlt: Improved validation of download option.

jowerner (approved):

Issue #138, opened 2021-09-21
Binary request bodies mixed up for OkHttp-based web connection


jowerner:
When sending binary content as part of a request body, the OkHttp-based web connection uploads the data all mixed up. 

HtmlUnit does not really support binary request bodies, so all binary request data must be passed as ISO 8859-1-encoded string only. Typically, the request carries a `Content-Type` header, for example `application/octet-stream` for binary data.

However, when building a string request body, OkHttp assumes UTF-8 if the content-type does not specify a charset. This is true for `application/octet-stream`, so the data gets encoded using UTF-8 which is wrong. We need to create a binary OkHttp request body for presumably binary request data.

PR #139, opened 2021-09-21
#138: Binary request bodies mixed up for OkHttp-based web connection

jowerner:
Create a binary OkHttp request body for presumably binary content.

### DIFF
--- a/src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/OkHttp3WebConnection.java
+++ b/src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/OkHttp3WebConnection.java
@@ -21,6 +21,7 @@ import java.net.InetSocketAddress;
 import java.net.Proxy;
 import java.net.URI;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.security.KeyManagementException;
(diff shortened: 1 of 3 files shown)


src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/OkHttp3WebConnection.java
```diff
@@ -206,7 +207,21 @@ protected Request createRequestWithStringBody(final URI uri, final WebRequest we
...
+        // create the request body
+        final RequestBody requestBody;
+
+        final MediaType mediaType = MediaType.get(contentType);
+        if (mediaType.charset() == null)
+        {
+            // interpret string body as binary/ISO_8859_1 content and create a binary request body
+            final byte[] bytes = body.getBytes(StandardCharsets.ISO_8859_1);
+            requestBody = RequestBody.create(mediaType, bytes);
+        }
```
h-arlt, line 219: Shouldn't we check whether the user requested to send the request's data as binary at all? Otherwise, we might end up in sending UTF-8 encoded text as (incorrectly decoded) binary stream.

For example, imagine a PUT request w/ content-type ''text/json" and UTF-8 encoded JSON string as body.

h-arlt (approved):

Issue #137, opened 2021-09-16
Improve cookie selection for OkHttp-based web connection


jowerner:
The cookie jar for the OkHttp client has to decide whether or not a cookie is to be sent back to the server for a certain URL. Instead of implementing this by ourselves, we should reuse the logic provided by OkHttp. This also includes things not covered so far, such as path matching.

PR #140, opened 2021-09-21
#137: Improve cookie selection for OkHttp-based web connection

jowerner:
- reuse logic from OkHttp to decide whether a cookie is to be sent back to a certain server
- handle host/wildcard domains consistently when converting cookies from/to HtmlUnit

### DIFF
--- a/src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/CookieJarImpl.java
+++ b/src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/CookieJarImpl.java
@@ -69,31 +69,8 @@ class CookieJarImpl implements CookieJar
     {
         cookieManager.clearExpired(new Date());
 
-        return cookieManager.getCookies().stream().filter(htmlUnitCookie -> matchesUri(htmlUnitCookie, url))
-                            .map(CookieJarImpl::toOkHttpCookie).collect(Collectors.toList());
-    }
-
-    /**
-     * Checks if the cookie is applicable for the URL.
-     *
-     * @param htmlUnitCookie
-     *            the cookie
-     * @param url
-     *            the URL
-     * @return whether or not the cookie is to be sent back to the URL
-     */
-    private static boolean matchesUri(final Cookie htmlUnitCookie, final HttpUrl url)
-    {
-        final String host = url.host();
-        String cookieDomain = htmlUnitCookie.getDomain();
-
-        // special cases
-        if (cookieDomain.equals("localhost.local"))
-        {
-            cookieDomain = "localhost";
-        }
-
-        return host.endsWith(cookieDomain);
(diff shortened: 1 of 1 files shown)


src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/CookieJarImpl.java
```diff
@@ -105,8 +82,11 @@ private static boolean matchesUri(final Cookie htmlUnitCookie, final HttpUrl url
      */
     private static Cookie toHtmlUnitCookie(final okhttp3.Cookie okHttpCookie)
     {
-        return new Cookie(okHttpCookie.domain(), okHttpCookie.name(), okHttpCookie.value(), okHttpCookie.path(),
-                          new Date(okHttpCookie.expiresAt()), okHttpCookie.secure(), okHttpCookie.httpOnly());
+        // add a leading dot if the domain is a "wildcard" domain -> ".example.com"
+        final String domain = okHttpCookie.hostOnly() ? okHttpCookie.domain() : "." + okHttpCookie.domain();
```
h-arlt, line 86: Are OkHTTP Cookies guaranteed to not start with a `.` character? If not, we'll have to use something like `StringUtils.prependIfMissing` in order to avoid cases when the domain of domain cookie already starts with a `.`.

src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/CookieJarImpl.java
```diff
@@ -120,12 +100,22 @@ private static Cookie toHtmlUnitCookie(final okhttp3.Cookie okHttpCookie)
...
-        builder.name(htmlUnitCookie.getName()).value(htmlUnitCookie.getValue()).domain(domain).path(htmlUnitCookie.getPath())
+        builder.name(htmlUnitCookie.getName()).value(htmlUnitCookie.getValue()).path(htmlUnitCookie.getPath())
                .expiresAt(htmlUnitCookie.getExpires().getTime());
 
+        final String domain = htmlUnitCookie.getDomain();
+        if (domain.startsWith("."))
+        {
+            // "wildcard" domain, such as ".example.com"
+            // the cookie builder expects the domain to be passed without the leading dot
+            builder.domain(StringUtils.stripStart(htmlUnitCookie.getDomain(), "."));
```
h-arlt, line 111:
```suggestion
            builder.domain(StringUtils.stripStart(domain, "."));
```

src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/CookieJarImpl.java
jowerner: Yep, OkHttp cookie domains never start with a `.`. To tell wildcard domains from host domains, the host-only flag is set to true/false accordingly.

Issue #162, opened 2021-11-17
Filtering capability in report syncs unrelated input fields


h-arlt:
XLT Load Test reports provide a filtering capability that can be used to narrow down the displayed table content. In case the report is customized with other input fields, those are synced with the current filter input value too.

To fix this, all relevant CSS selectors used in the linked JavaScript file should be revisited and adjusted properly so that unrelated input fields won't match and stay untouched.

PR #163, opened 2021-11-17
GH #162: Filtering capability in report syncs unrelated input fields

h-arlt:
CSS selector adjustments to target our own filter input only when syncing and/or (re)focusing and custom input fields stay untouched.


jowerner (approved):
Nice!

PR #174, opened 2021-12-16
57: Hide credentials passed in the URL in the load test report

h-arlt:
- introduce new property to control whether or not to remove user-info from request URL
- remove user-info from request URL if need to do so before writing timer data to CSV
- added tests for new method 'removeUserInfo' of utility class 'com.xceptance.xlt.engine.util.UrlUtils'
- added documentation of new property to all 'default.properties' files in sample test suites

### DIFF
--- a/src/main/java/com/xceptance/xlt/engine/XltHttpWebConnection.java
+++ b/src/main/java/com/xceptance/xlt/engine/XltHttpWebConnection.java
@@ -16,6 +16,7 @@
 package com.xceptance.xlt.engine;
 
 import java.io.IOException;
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
@@ -314,7 +315,7 @@ public class XltHttpWebConnection extends CachingHttpWebConnection
             if (logger.isInfoEnabled())
             {
                 logger.info(response.getWebRequest().getHttpMethod().name() + " - " + response.getStatusCode() + " - " + runTime +
-                            " ms - " + response.getWebRequest().getUrl() + " " + response.getWebRequest().getRequestParameters());
+                            " ms - " + removeUserInfoIfNecessary(response.getWebRequest().getUrl()) + " " + response.getWebRequest().getRequestParameters());
             }
 
             // set response data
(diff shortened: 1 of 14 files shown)


src/main/java/com/xceptance/xlt/engine/util/UrlUtils.java
```diff
@@ -223,4 +225,37 @@ public static String getUrlEncodedParameters(final List<NameValuePair> parameter
...
+            }
+            if (info.getFragment() != null)
+            {
+                sb.append('#').append(info.getFragment());
+            }
+            return sb.toString();
+        }
+
+        return null;
+    }
```
jowerner, line 260: Small performance optimization: We often have URL objects, not strings. What about providing a second version of `UrlUtils.removeUserInfo()` that takes a URL and builds the resulting user-less url string from the already parsed components in URL? This way we don't need to convert the URL to a string and afterwards parse the string again into its pieces.

src/main/java/com/xceptance/xlt/engine/XltHttpWebConnection.java
```diff
@@ -486,4 +486,17 @@ protected static void putAdditionalRequestData(RequestData requestData, WebReque
...
+
+        // remove user-info from request URL if we need to (GH #57)
+        if(SessionImpl.REMOVE_USERINFO_FROM_REQUEST_URL)
+        {
+            urlString = UrlUtils.removeUserInfo(urlString);
+        }
+
+        return urlString;
+    }
 }
```
jowerner, line 502: In the result browser, the user info is still present in the URLs. Maybe we can fix this by removing the user info in `XltHttpWebConnection.cloneWebRequest()`?

h-arlt: Done.

However, I'm not absolutely sure that this is a good idea as the result browser is primarily used for debugging and analysis purposes and users might wonder why this information is removed from there as well. As for me, I'd expect the result browser to represent the data that actually transmitted.

src/main/java/com/xceptance/xlt/engine/util/UrlUtils.java
h-arlt: Done.

src/main/java/com/xceptance/xlt/engine/XltHttpWebConnection.java
jowerner: Result browsers are often linked to in load test reports. If these reports are shared carelessly with other people, they might see this info again. So better safe than sorry.

jowerner (approved):

PR #188, opened 2022-01-20
Remove xlt docs

jowerner:


README.md
```diff
@@ -1,26 +1,24 @@
...
-XLT has been open sourced on 30 January 2020 under the Apache License 2.0. Stay tuned for more!
+XLT has been open sourced on 30 January 2020 under the Apache License 2.0.
 
 # How to Build XLT
 
 The XLT build process is based on the **Apache Ant** build tool. The build works best when being run on a Linux machine.
 
 ## Required Tools
 
-* Latest JDK 8 (or 11)
```
DerNamenlose, line 15: Was support for building on JDK 11 removed?

jowerner: No, but we do not test that regularly since we build with Java 8 only. Proper Java 11 support comes (back) with the other merge request.

PR #194, opened 2022-01-26
Update log4j

jowerner:


h-arlt (commented):
Looks good to me. However, the dev-mode lo4j2 configurations differ to the former ones w/ respect to max file size for roll-over. Was this change intentionally?

jowerner:
> Looks good to me. However, the dev-mode lo4j2 configurations differ to the former ones w/ respect to max file size for roll-over. Was this change intentionally?

Hawkeye! No, this is a remnant from when I tested the roll-over stuff and got committed by mistake. Fixed. Thx!

Issue #208, opened 2022-03-08
"Unexpected end of ZLIB input stream" error when downloading results


jowerner:
When downloading results from an agent controller, the master controller may sometimes fail with this error:

```
Downloading test results... Please be patient, it might take some time...
    0% ... 10% ... 20% ... 30% ... 40% ... 50% ... 60% ... 70% ... 80% ... - FAILED!

 -> Agent Controller ac001 <https://localhost:8500> - Unexpected end of ZLIB input stream
```

The `mastercontroller.log` has some details:

```
java.io.EOFException: Unexpected end of ZLIB input stream
        at java.util.zip.InflaterInputStream.fill(Unknown Source) ~[?:?]
        at java.util.zip.InflaterInputStream.read(Unknown Source) ~[?:?]
        at java.util.zip.ZipInputStream.read(Unknown Source) ~[?:?]
        at java.io.FilterInputStream.read(Unknown Source) ~[?:?]
        at org.apache.commons.io.IOUtils.copyLarge(IOUtils.java:1158) ~[commons-io-2.8.0.jar:2.8.0]
        at org.apache.commons.io.IOUtils.copy(IOUtils.java:878) ~[commons-io-2.8.0.jar:2.8.0]
        at org.apache.commons.io.IOUtils.copyLarge(IOUtils.java:1135) ~[commons-io-2.8.0.jar:2.8.0]
        at org.apache.commons.io.IOUtils.copy(IOUtils.java:854) ~[commons-io-2.8.0.jar:2.8.0]
        at com.xceptance.common.util.zip.ZipUtils.unzipFile(ZipUtils.java:372) ~[xlt-6.0.0.jar:6.0.0]
        at com.xceptance.xlt.mastercontroller.ResultDownloader.downloadTestResults(ResultDownloader.java:530) ~[xlt-6.0.0.jar:6.0.0]
        at com.xceptance.xlt.mastercontroller.ResultDownloader$4.call(ResultDownloader.java:404) ~[xlt-6.0.0.jar:6.0.0]
        at com.xceptance.xlt.mastercontroller.Poll$1.run(Poll.java:91) [xlt-6.0.0.jar:6.0.0]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) [?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) [?:?]
        at java.lang.Thread.run(Unknown Source) [?:?]
```

[...]

PR #209, opened 2022-03-08
#208: "Unexpected end of ZLIB input stream" error when downloading results

jowerner:
- add content length header manually as the servlet API method allows int values only

### DIFF
--- a/src/main/java/com/xceptance/xlt/agentcontroller/FileManagerServlet.java
+++ b/src/main/java/com/xceptance/xlt/agentcontroller/FileManagerServlet.java
@@ -102,7 +102,7 @@ public class FileManagerServlet extends HttpServlet
                 in = new FileInputStream(file);
 
                 resp.setStatus(HttpServletResponse.SC_OK);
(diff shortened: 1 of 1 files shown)


src/main/java/com/xceptance/xlt/agentcontroller/FileManagerServlet.java
```diff
@@ -101,14 +101,12 @@ protected void doGet(final HttpServletRequest req, final HttpServletResponse res
                 final File file = new File(rootDirectory, fileName);
                 in = new FileInputStream(file);
 
-                resp.setContentLength((int) file.length());
-                // resp.setContentType("???");
-
+                resp.setStatus(HttpServletResponse.SC_OK);
+                resp.setHeader("Content-Length", String.valueOf(file.length()));
```
DerNamenlose, line 105: I'd prefer using `resp.setContentLengthLong(file.length());` instead.

jowerner: Now that I know that this method exists, I'd prefer that too.   ;-)
Done.

Issue #73, opened 2021-02-05
Extend Merge Rules by Method


rschwietzke:
Extend the merge rules by method to deal with GET, POST, and more easily. In addition, rework the switch that produces method tracking to track only the method without the overhead of POST data for instance.

PR #217, opened 2022-03-24
#73: Extend Merge Rules by Method

jowerner:
- added a new request filter type to allow request processing by HTTP method
- adjusted/extended unit tests
- added new filter pattern to project.properties


samples/testsuite-performance/config/project.properties
```diff
@@ -82,6 +82,7 @@ com.xceptance.xlt.loadtests.TLWSimpleURL.class = com.xceptance.xlt.performance.t
 #   contentTypePattern [c] ... Reg-ex defining a matching response content type
 #   statusCodePattern [s] .... Reg-ex defining a matching status code
 #   urlPattern [u] ........... Reg-ex defining a matching request URL
+#   httpMethodPattern [m] .... Reg-ex defining a matching HTTP method
```
h-arlt, line 85: Do we want to shorten this property to `methodPattern` instead (so that users can link it with the appropriate category key `m` more easily)? IMO the prefix 'http' is not really necessary in this context (similar to `statusCodePattern`).

jowerner: Yup, done.

Issue #237, opened 2022-07-06
Resultbrowser: refactor image loading


jeanNimb:
It should be checked how the the page load costs can be reduced.

some ideas:
* replace all images with an SVG that are inline
* forget the spinner gif and make it a static
* remove the images dir

PR #242, opened 2022-07-28
237 resultbrowser refactor image loading

jeanNimb:
[Issue](https://github.com/Xceptance/XLT/issues/237)


resultbrowser/src/css/default.css
```diff
@@ -50,11 +51,11 @@ body {
...
-    background-image: url(data-url:../images/progress.gif);
-    width: 24px;
-    height: 24px;
-    float: right;
-    margin: 2px;
+    margin: 10px;
+    position: fixed;
+    top: 0;
+    right: 0;
+    z-index: 2147483647;
```
h-arlt, line 58: As we have full control over our CSS, there is no need to use such a high value for z-index. IMO, a value of 10 should be sufficient.

resultbrowser/src/css/default.css
```diff
@@ -81,7 +82,7 @@ body {
     background-color: white;
 }
 
-#menu > ul > li {
+#menu>ul>li {
```
h-arlt, line 85: Please revert this change and all the others that removed the spaces from the child combinator. It's really hard to read selectors formatted like this.

resultbrowser/src/css/default.css
```diff
@@ -595,3 +606,30 @@ table.key-value-table td.value .csep {
...
+  /* Safari */
+  @-webkit-keyframes spin {
+    0% { -webkit-transform: rotate(0deg); }
+    100% { -webkit-transform: rotate(360deg); }
+  }
+  
+  @keyframes spin {
+    0% { transform: rotate(0deg); }
+    100% { transform: rotate(360deg); }
+  }
```
h-arlt, line 635: Please do not remove the final newline character at end-of-file.

h-arlt (changes requested):
See comments inline.

resultbrowser/src/css/default.css
jeanNimb: Done

resultbrowser/src/css/default.css
jeanNimb: Done

resultbrowser/src/css/default.css
jeanNimb: Done

h-arlt (approved):

PR #243, opened 2022-08-05
remediate vulnerabilities

hknochenhauer:


jowerner:
Thanks for providing this PR! Please note that I had to revert the updates for `htmlunit-core-js` and `progressbar` because of compile or functional issues. I also made some additional minor tweaks.

Regarding the change you made to `resultbrowser/package.json`: Is this a local change that slipped into this PR by accident or do you think that `@parcel/css` is missing?

Issue #240, opened 2022-07-08
Support alternative logger configuration file formats


jowerner:
When looking for a log4j2 configuration file in the test suite, an XLT agent tries `config/log4j2.properties` **only**. However, users may wish to use another supported format, e.g. XML, JSON or YAML. XLT should support these alternative formats as well. Can be done by specifying multiple target files in property `log4j2.configurationFile`.

**Workaround:**

Add `-Dlog4j2.configurationFile=config/log4j2.xml` or similar to `config/jvmargs.cfg` to override the defaults.

jowerner:
Note that listing multiple configuration files in property `log4j2.configurationFile` has the side effect that all these files are loaded (should they exist) and their configurations merged. In order to avoid unexpected results, decide on the file format (say XML), create the respective configuration file (say `log4j2.xml`) and delete all other `log4j2.*` files.

jowerner:
Support for YAML and JSON in log4j2 relies on several 3rd-party libraries. In order to keep the footprint of XLT smaller, we will for now support XML only and wait for feedback.

PR #245, opened 2022-08-11
#240: Support alternative logger configuration file formats

jowerner:
Extended property `log4j2.configurationFile` to look for log4j2.yaml, log4j2.json, and log4j2.xml as well.


jowerner:
Note that listing multiple configuration files in property `log4j2.configurationFile` has the side effect that all these files are loaded (should they exist) and their configurations merged. In order to avoid unexpected results, decide on the file format (say XML), create the respective configuration file (say `log4j2.xml`) and delete all other `log4j2.*` files.

h-arlt:
Changes look good to me. However, YAML requires additional libraries that are declared as optional dependency by log4j-core

And, after downloading load test results the `config` directory misses the file `log4j2.yaml` file, and `log4j2.xml` respectively.

jowerner:
Removed support for YAML annd JSON for now. Let's wait for feedback. The log4j2.xml file should now be copied to the results just fine.

h-arlt (approved):

PR #255, opened 2022-09-01
fix vulnerabilities

hknochenhauer:


rschwietzke:
Due to many conflicts and failures due to the lib dependency choices, we will close this and open an extra one for JQuery

Issue #268, opened 2022-10-24
Update list of supported data centers


jowerner:
Amazon and Google both opened new data centers recently. We need to add them to `ec2_admin` and `gce_admin`, respectively.

PR #270, opened 2022-10-24
#268: Update list of supported data centers

jowerner:


h-arlt:
We might also add the following data-centers:

AWS
* ap-south-2 - Asia Pacific, Hyderabad
* eu-south-2 - Europe, Spain
* eu-central-2 - Europe, Zurich

src/main/java/com/xceptance/xlt/gce/GceAdminUtils.java
h-arlt, line 1: It is really _North Virginia_ or should it be just _Virginia_? And, should we just display _Texas_ instead of _Dallas_?

h-arlt (approved):

src/main/java/com/xceptance/xlt/gce/GceAdminUtils.java
rschwietzke: Google calls it N. Virginia and Dallas. It is a mix of Countries, States and cities.
https://cloud.google.com/about/locations#lightbox-regions-map

h-arlt: I used https://cloud.google.com/compute/docs/regions-zones/ as reference which uses a consistent naming scheme (in contrast to the link you posted).
As we used the name of the state for the other US locations, I'd stay with Texas instead of Dallas. Similar for North Virginia; this is no valid name of a US  state but Virginia is.

PR #287, opened 2022-12-05
#285: Update to latest Selenium/HtmlUnit driver

### LINKED ISSUE #285: Update to latest Selenium/HtmlUnit driver
Selenium 4.7.0 and a matching HtmlUnit driver are now available.

jowerner:
Closes #285


h-arlt:
Changes look good to me.

However, `commons-codec`  - a transitive dependency of HtmlUnit and HttpClient - may be updated from `1.11`  to `1.15` (the version used by HU).

src/main/java/com/xceptance/xlt/engine/util/DefaultWebDriverFactory.java
```diff
@@ -72,8 +69,7 @@
      xlt,
      firefox_clientperformance,
      chrome_clientperformance,
-     edge,
-     opera
+     edge
```
h-arlt, line 72: Do we really want to drop support for Opera entirely? Or do want to add support for Chromium-based browsers such as recent Opera and Edge? Last but not least, is explicit support for Edge still necessary?

h-arlt (approved):

src/main/java/com/xceptance/xlt/engine/util/DefaultWebDriverFactory.java
h-arlt: Dedicated driver server for recent Opera can be downloaded [here](https://github.com/operasoftware/operachromiumdriver/releases)

jowerner:
> Changes look good to me.
> 
> However, `commons-codec` - a transitive dependency of HtmlUnit and HttpClient - may be updated from `1.11` to `1.15` (the version used by HU).

Yep, missed that during HU update. Done.

src/main/java/com/xceptance/xlt/engine/util/DefaultWebDriverFactory.java
h-arlt: I'm okay for the Edge case ;) However, there might be people who want to run tests with Chrome/Chromium and Opera simultaneously; that's not possible any longer unless they provide their own WebDriver instance.

What I'm trying to say is: if Opera can be used by simply using the ChromiumDriver (not ChromeDriver) and the opera driver server, we might why not supporting that. OperaDriver might have been removed from Selenium because is became obsolete due to ChromiumDriver.

Issue #309, opened 2023-01-13
Enable the average report to display more than one average


rschwietzke:
![image](https://user-images.githubusercontent.com/1793856/212322583-89341594-4bbc-439e-9b1c-ac2e2a37b6c8.png)

To enable more control over the averages displayed, we want to have the ability to configure different moving averages and display them in the same chart, such as:

* last 5% of the data
* last 5 min
* last 5 requests

The 5 here is of course just an example. Besides that different type, you should be able to have more than one of the average displayed, such as last 5 min and last 1 min for instance.

rschwietzke:
We have to take care of the properties in a meaningful way, here is the PO suggestion.

## Base
We have now two types of averages. The one that is in every chart and the averages on the average charts.

### Common Average
![image](https://user-images.githubusercontent.com/1793856/219391647-96058b8c-061d-4192-a711-51fa64de6adf.png)

### Average Chart
![image](https://user-images.githubusercontent.com/1793856/219391765-5a859292-e31c-432c-9186-d5ff85cff05b.png)

Currently, the moving average for all charts is set like that:

```
## The percentage of values taken when calculating the moving average series.
com.xceptance.xlt.reportgenerator.charts.movingAverage.percentageOfValues = 5
``` 

## New Properties 
We have to distinguish between the common average and the average chart. The new averages support the following line and data types: percentage, time window (values in the past time window from this current value), and count (last X values)

### Common Average

``` 
## What type to use
com.xceptance.xlt.reportgenerator.charts.commonAverage.type = percentage|time|count
## What value to set, this 5%, 5 min, count = 5, the latter is always based on what type the average reflects, such as 5 requests, 5 actions, 5 transactions, and so forth. The time value supports our common syntax for time specs, such as 5s 5m 5h or 1:30... Count doesn't require a unit. Percentage uses % without a space.
com.xceptance.xlt.reportgenerator.charts.commonAverage.value = 5%|5m|5
``` 

### Average Chart
We will use an index for the charts and this is 1 to 10.  Might not make sense to have that many, but if one wants to do that. 

``` 
## What type to use
com.xceptance.xlt.reportgenerator.charts.commonAverage.type = percentage|time|count
## What value to set, this 5%, 5 min, count = 5, the latter is always based on what type the average reflects, such as 5 requests, 5 actions, 5 transactions, and so forth. The time value supports our common syntax for time specs, such as 5s 5m 5h or 1:30... Count doesn't require a unit. Percentage uses % without a space.
[...]

PR #320, opened 2023-01-27
#309 Enable the average report to display more than one average

Glenmorangie:


config/reportgenerator.properties
```diff
@@ -86,6 +86,35 @@ com.xceptance.xlt.reportgenerator.charts.height = 300
...
+# When enabled this will draw additional moving averages in the same average chart.
+# The values must be indexed and need a certain format.
+#
+# Format:
+# 
+#   com.xceptance.xlt.reportgenerator.charts.movingAverage.additional.<index>.<percentile|time|requests>
+# 
+#   index ........ Numeric value for indexing.
+#
+#   percentile ... The percentage of values taken when calculating the moving average series.
```
h-arlt, line 102: When a percentage of a value set is used, we should name it `percentage`.

src/main/java/com/xceptance/xlt/report/util/JFreeChartUtils.java
```diff
@@ -935,6 +1004,88 @@ public static TimeSeries createMovingAverageTimeSeries(final TimeSeries series,
         // return MovingAverage.createMovingAverage(series, avgSeriesName, samples, samples);
```
h-arlt, line 1004: Is this still needed or can this line be removed?

src/main/java/com/xceptance/xlt/report/util/MovingArerage.java
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * Hold the data for additional moving averages.
+ */
+public class MovingArerage
+{
+    /**
+     * Additional moving averages. 
+     */
+    public enum MovingAverages
```
h-arlt, line 26: What about to rename this enum to something that better fits its intention, e.g. `MovingAverageType`?

src/main/java/com/xceptance/xlt/report/util/MovingArerage.java
```diff
@@ -0,0 +1,134 @@
...
+    
+    /**
+     * Given property value for the time. 
+     */
+    private String timeString = "";
+    
+    /**
+     * Current used type for moving average.
+     */
+    private MovingAverages type;
```
h-arlt, line 62: In case the values are not expected to change (as the absence of any setter method implies) once they have been set via c'tor, they should be declared as `final`. 
And, there is no need to initialize fields with Java defaults (`int` and `long` are initialized implicitly w/ 0).

BTW do we really need two fields here? IMO having one field of type `long` would suffice too.

src/main/java/com/xceptance/xlt/report/util/JFreeChartUtils.java
```diff
@@ -935,6 +1004,88 @@ public static TimeSeries createMovingAverageTimeSeries(final TimeSeries series,
...
+     * @param requestCount
+     *            overall amount of requests
+     * @return the time series
+     */
+    public static TimeSeries createMovingAverageTimeSeriesAdditonalRequests(final TimeSeries series, final int value, final double requestCount)
+    {
+        // percentage value from overall requests
+        int percentage = (int) Math.floor((value / requestCount) * 100);
+        // samples for chart data items
+        final int samples = Math.max(2, series.getItemCount() * percentage / 100);
```
h-arlt, line 1081: This doesn't seem to be correct as it uses a percentage of values instead of a fixed amount.  Maybe, I just don't get your intention here. But I'd expect something like the following
```
        final int samples = Math.max(2, value);
```

We might also have to verify that `value` is not greater than `requestCount`.

src/main/java/com/xceptance/xlt/report/util/JFreeChartUtils.java
```diff
@@ -935,6 +1004,88 @@ public static TimeSeries createMovingAverageTimeSeries(final TimeSeries series,
...
+        DecimalFormat decimalFormat = new DecimalFormat("0.##E0");
```
h-arlt, line 1082: What about to made this a static field such it won't be created over and over again?

src/test/java/com/xceptance/xlt/report/providers/DummyReportProviderConfiguration.java
```diff
@@ -88,6 +91,12 @@ public int getMovingAveragePercentage()
     {
         return 0;
     }
+    
+    @Override
+    public List<MovingArerage> getAdditonalMovingAverages()
+    {
+        return Collections.emptyList();
```
h-arlt, line 98: Since Java 9, List interface provides method `of` that returns type-safe immutable collections including a stable reference to empty list. Thus, there is no need for `Collections.emptyList()` anymore.

```suggestion
        return List.of();
```

This is no CR, just FYI.

src/main/java/com/xceptance/xlt/report/util/JFreeChartUtils.java
```diff
@@ -85,7 +86,9 @@
         /**
          * blue, ...
          */
-        public static final ColorSet AVERAGES = new ColorSet(COLOR_MOVING_AVERAGE, COLOR_MEDIAN, COLOR_MEAN);
+        public static final ColorSet AVERAGES = new ColorSet(COLOR_MOVING_AVERAGE, COLOR_MEDIAN, COLOR_MEAN, COLOR_MOVING_AVERAGE_ADD_1,
```
h-arlt, line 89: To clarify: what if user configured more moving averages than colors available?

h-arlt (changes requested):

config/reportgenerator.properties
```diff
@@ -86,6 +86,35 @@ com.xceptance.xlt.reportgenerator.charts.height = 300
 ## The percentage of values taken when calculating the moving average series.
 com.xceptance.xlt.reportgenerator.charts.movingAverage.percentageOfValues = 5
 
+###############################################################################
+#
+# Additional Moving Average Settings
+#
```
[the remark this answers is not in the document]
h-arlt: Had a talk w/ the PO about that topic and he was fine w/ the current approach as the "default" moving average is used for all charts whereas the additional moving averages are used on the Averages Charts only.

Of course, consolidation of the two formats might still be an option (although some way to "mark" a moving average as default would be needed then).

@jowerner What do you think?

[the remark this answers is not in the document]
jowerner: See [here](https://github.com/Xceptance/XLT/issues/309#issuecomment-1433774327) for more input regarding this topic.

Issue #291, opened 2022-12-12
Move summary chart above data tables


rschwietzke:
To make the summary chart more easily accessible, especially when the tables (requests) are very long, we should move the Summary section with the chart above the table with all details. The individual charts stay under the table. 

![image](https://user-images.githubusercontent.com/1793856/207003722-d4b19381-e1e8-4e80-9e47-fd5e4f291843.png)

js-xc:
While I appreciate this as a learning experience I wonder whether the change will be useful for load testers. I understand that it might be valuable for non-tech users to see the overview graphic first, but for evaluating the report I usually need the data tables more often.

rschwietzke:
It is about seeing potential problems early and because we share that with customers, they should see that first. Additionally, very long tables take long to load and so the chart is already there.

PR #321, opened 2023-01-31
#291: move summary chart above data tables

js-xc:
While I appreciate this as a learning experience I wonder whether the change will be useful for load testers. I understand that it might be valuable for non-tech users to see the overview graphic first, but for evaluating the report I usually need the data tables more often.

### DIFF
--- a/config/xsl/loadreport/util/timer-section.xsl
+++ b/config/xsl/loadreport/util/timer-section.xsl
@@ -8,7 +8,6 @@
         <xsl:param name="runtimeIntervalsNode"/>
         <xsl:param name="type"/>
         
-        <!-- <h3 class="no-print">Summary</h3> -->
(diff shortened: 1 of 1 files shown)


config/xsl/loadreport/util/timer-section.xsl
```diff
@@ -7,6 +7,17 @@
         <xsl:param name="directory"/>
         <xsl:param name="runtimeIntervalsNode"/>
         <xsl:param name="type"/>
+        
+        <h3 class="no-print">Summary</h3>
```
jowerner, line 11: According to feedback, the Summary header is not really needed and should be removed.

jowerner: Please remove the heading altogether. No need to have it as comment in the template.

PR #329, opened 2023-02-07
322 add agent summary chart

js-xc:


jowerner (changes requested):
The template works as expected, however, I would prefer to not copy the XSL instructions needed for embedding the summary charts. Similar to `util/timer-chart.xsl`, please invent a reusable `util/agent-chart.xsl` sub-template and call it for the summary chart area and each individual agent chart area.

PR #336, opened 2023-02-15
108 add ip statistics

js-xc:
We're actually reporting the first IP recorded for each request, which makes only very limited sense (not necessarily the one that has been used in the end?). Would need better input data for IPs that were used (most request data uses cached connections, i.e. has no IP data at all).


rschwietzke:
Please rebase against `develop` which is v7.0 and resubmit.

PR #342, opened 2023-02-21
335 gradle classpath for kotlin

js-xc:


jowerner (changes requested):
Please make similar changes to the `agent.cmd` file.

PR #347, opened 2023-02-28
311 report header

js-xc:


rschwietzke:
Looks good. I would like to suggest using an animation CSS transition to make the shrinking a smoother.

rschwietzke:
You are right, that does not look good. Let's keep it un-animated for now and we take care of that later. Can you please compare to the old branch 323-chart-refresh state? Please, seems that header and logo list some height.

image

Left this branch, right - the original branch for the redone report.

rschwietzke:
Okidoki.

rschwietzke:
The original had more bottom padding to fluff it up, but let's go with the current state.

jowerner:
If sticky, the header seems to obstruct the menu bar by some pixels. When fixing this, please keep in mind that the menu bar has two different heights, one for smaller and one for larger screens.

jowerner:
In the non-XTC case, don't show "Created with ..." in the sticky header, but "Performance Test Report", or "\<Project Name\> - Performance Test Report" if a project name was configured. Adjust the font size as appropriate.

jowerner:
When displaying the XTC infos as key/value pairs, make the values bold, not the keys, to put the focus on them.

config/xsl/common/sections/footer.xsl
```diff
@@ -2,9 +2,21 @@
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
 
 <xsl:template name="footer">
-    <footer>
-        Copyright (c) 2005-2023 <a href="http://www.xceptance.com/" title="Visit the Xceptance website">Xceptance Software Technologies GmbH</a>
-    </footer>
+	<xsl:param name="productUrl" select="/testreport/configuration/version/productURL" />
+    <xsl:param name="productName" select="/testreport/configuration/version/productName" />
+    <xsl:param name="productVersion" select="/testreport/configuration/version/version" />
+    <div id="footer">
```
jowerner, line 8: Please keep using the `<footer>` tag instead of `div`.

config/xsl/common/sections/footer.xsl
```diff
@@ -2,9 +2,21 @@
...
+        Copyright (c) 2005-2023 <a href="http://www.xceptance.com/" title="Visit the Xceptance website">Xceptance Software Technologies GmbH</a><br />
+        Created with
+                <a href="{$productUrl}?piwik_campaign=TestReport">
```
jowerner, line 11: In the header, we use `<a href="{$productUrl}?source=TestReport">` as the URL. Use that one here as well.

config/testreport/css/default.css
```diff
@@ -650,9 +650,57 @@ nav
...
                 border-bottom: 1px solid;
             }
-        #header .productversion {
+        #header .productversion, #footer .productversion {
             padding-left: 0.25rem;
         }
+        
+        #header div.ltinfo {
+        	display: inline; 
+        	margin-right: 3ch;
```
jowerner, line 659: PO wants less space between key/value pairs.

jowerner (changes requested):
See comments in-line and on the Conversation tab.

config/xsl/common/sections/footer.xsl
js-xc: I did not even change this (someone else did that recently, I got it with rebase), but of course I can do that.

js-xc: Done.

config/xsl/common/sections/footer.xsl
js-xc: Done.

js-xc:
All done (?), ready for next round of reviews... (If no project name was set - for non-XTC-reports - the small header will just display "Performance Test Report".)

jowerner (approved):

PR #349, opened 2023-03-07
added request methods section to report

js-xc:


src/main/java/com/xceptance/xlt/report/providers/RequestMethodsReportProvider.java
```diff
@@ -0,0 +1,73 @@
...
+import com.xceptance.xlt.api.engine.RequestData;
+import com.xceptance.xlt.api.report.AbstractReportProvider;
+
+/**
+ * 
+ */
+public class RequestMethodsReportProvider extends AbstractReportProvider
+{
+    /**
+     * A mapping from request methods to their corresponding {@link ResponseCodeReport} objects.
```
jowerner, line 32: Fix javadoc: `ResponseCodeReport` -> `RequestMethodReport`

jowerner:
Please put the new *HTTP Request Methods* sections before the *HTTP Response Codes section*. I have no real reason for that, just the feeling that request stuff should come before response stuff.

jowerner:
Don't forget to add a corresponding entry to the *Network* menu in the top-level menu bar. (BTW, clicking those menu entries leads to a weird scrolling behavior at the moment, but that's a different story.)

jowerner (changes requested):
See comments in-line and on the Conversation tab.

js-xc:
I think the weird scrolling behavior is caused by the jQuery "scrolling magic" (quote from the actual code), but I don't really get what this is doing and why.

jowerner (approved):

Issue #306, opened 2023-01-13
Make XLT reports responsive


rschwietzke:
Currently, XLT reports are not fully responsive. Especially the menu bar is not behaving properly. Introduce a menu bar that becomes a hamburger menu when the screen size is too small. 

Check all the other areas as well and see where we can make it support a smaller devices better. 

It is properly recommended to keep tables the way they are and just ensure that they can be zoomed in on as well as scrolled horizontally.

rschwietzke:
#327 already delivers a better navigation menu and responsive charts. The hamburger menu is open.

rschwietzke:
The hamburger menu has to be done without JS. Just google for it, there are solutions such as: https://www.google.com/search?q=hamburger+menu+pure+css

PR #353, opened 2023-03-15
#306 - Add hamburger menu

FabianHeydrich:
see #306 as a reference


rschwietzke (changes requested):
Please see typical breaking points. 
https://testsigma.com/blog/css-breakpoints/

I am not sure what I used for the current change from large to small font... but one is likely a typicaly breakpoint.

rschwietzke:
Besides that breakpoint feature, it looks ok. Please have the menu on the left side, also, we updated the menu too in a later XLT version, so it does not use superfish anymore and has different CSS, you might have to rebase your branch first.

FabianHeydrich:
@rschwietzke I added breakpoints for the different screen sizes. Still open is the resizing of the Load Profile table. From waht I have seen so far the size of that table is fixed set in the HTML code

```
style="width: 1079.35px;"
```

rschwietzke:
I will check and first rebase with the latest report state.

rschwietzke:
What you see is the behavior of the table.js. There is not responsive version of a table, it will, sooner or later, turn into something horizontally scrollable.

FabianHeydrich:
After the rebase I need to apply some changes to bring the hamburger menu back

PR #357, opened 2023-03-27
Htmlunit 3

jowerner:
Updates HtmlUnit to version 3.0.0. Also updates HtmlUnit driver, etc.

**ATTENTION:** Please don't squash-merge this pull request. Use a regular merge instead. We need to keep track that the changes originally came from the htmlunit-integration branch.


h-arlt:
Most source files use 2023 as end of copyright but some still use the "old" value (2022).

See sources in package `com.xceptance.xlt.engine.xltdriver` and below.

jowerner:
> Most of the affected source files use 2023 as end of copyright but some still use the "old" value (2022).
> 
> See sources in package `com.xceptance.xlt.engine.xltdriver` and below.

Fixed.

h-arlt (approved):

PR #365, opened 2023-04-05
307 sorting and filtering via hash in online report feature updated

xceptance-dan:
Rebased against developed. Also added the feature to remove an anchor when the backToTop handler comes into play.


js-xc:
I think we need to rebase with `develop` again. No conflicts, but some updates: 
* for me there was an issue with `th colspan` on requests page (see `/config/xsl/loadreport/util/timer-table.xsl`:line38ff, there should be a different colspan for type request (5) and other types (4)
* the *web vitals* page that has been recently added has no `th ids` yet, so sorting url param will be empty/not working (i.e. `#=desc`)
* not sure whether diff report/trend report were officially part of the task, the functionality is basically there but as they also don't have `th ids` yet sorting url params will be empty, too

Issue #372, opened 2023-05-24
Read properties files as UTF-8


jowerner:
At the moment, XLT reads properties files using the ISO-8859-1 encoding. This is for historical reasons only and the only case where UTF-8 is not used yet. Time to go with UTF-8 for properties files as well. This should also simplify maintaining properties files with non-Western text content. No need to deal with unicode escapes any longer!

Note that this is another breaking change that needs to be documented in the migration guide for XLT 7.

PR #373, opened 2023-05-24
#372: Read properties files as UTF-8

jowerner:
- introduced Utf8Reader incl. test case
- use Utf8Reader when reading properties
- make sure we use UTF-8 when writing or appending text to properties files

### DIFF
--- a/src/main/java/com/xceptance/common/io/Utf8Reader.java
+++ b/src/main/java/com/xceptance/common/io/Utf8Reader.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) 2005-2023 Xceptance Software Technologies GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.xceptance.common.io;
+
+import java.io.CharArrayReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.Reader;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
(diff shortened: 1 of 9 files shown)


src/main/java/com/xceptance/common/io/Utf8Reader.java
```diff
@@ -0,0 +1,104 @@
...
+import java.io.Reader;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * A {@link Reader} implementation that reads UTF-8-encoded text from an {@link InputStream} throwing an
+ * {@link IOException} if the text is not valid UTF-8.
```
h-arlt, line 30: It might be worth to add a few words how this class differs to Java's InputStreamReader using UTF-8 charset. WDYT?

src/main/java/com/xceptance/common/io/Utf8Reader.java
```diff
@@ -0,0 +1,104 @@
...
+        try
+        {
+            chars = getAsChars(bytes, StandardCharsets.UTF_8);
+        }
+        catch (final CharacterCodingException cce)
+        {
+            throw new IOException("Data does not represent UTF-8-encoded text", cce);
+        }
+
+        charArrayReader = new CharArrayReader(chars);
```
h-arlt, line 63: Instance field is only initialized when reading all bytes did succeed. IMO, the following makes the control flow more clear:
```suggestion
        
        try
        {
            final char[] chars = getAsChars(bytes, StandardCharsets.UTF_8);
            charArrayReader = new CharArrayReader(chars);
        }
        catch (final CharacterCodingException cce)
        {
            throw new IOException("Data does not represent UTF-8-encoded text", cce);
        }
```

h-arlt (approved):

rschwietzke:
Seems that this is all done already with Java 9.

UTF-8 Properties Files
In Java SE 9, properties files are loaded in UTF-8 encoding. In previous releases, ISO-8859-1 encoding was used for loading property resource bundles. UTF-8 is a much more convenient way to represent non-Latin characters.

Most existing properties files should not be affected: UTF-8 and ISO-8859-1 have the same encoding for ASCII characters, and human-readable non-ASCII ISO-8859-1 encoding is not valid UTF-8. If an invalid UTF-8 byte sequence is detected, the Java runtime automatically rereads the file in ISO-8859-1.

PR #384, opened 2023-07-06
339 mark responses for skipped requests uniquely

EliasTatX:
adjusted naming and comment for fake response body and header and added a new distinguishing header unique to those responses


jowerner:
Branch was merged outside of GitHub.

PR #405, opened 2023-08-14
Improved parser with less exposure to the JIT problem, but still not …

rschwietzke:
…fully free of it, but mostly really fast


rschwietzke:
Replaced by #418

PR #418, opened 2023-09-15
New CSV Parser and removed it from data classes

rschwietzke:
Faster parser. Less exposure to JIT problem.


src/test/java/com/xceptance/common/util/CsvLineDecoderTest.java
```diff
@@ -0,0 +1,374 @@
...
+            return;
+        }
+        fail("No exception was raised");
+    }
+
+    @Test
+    public void empty()
+    {
+        var r = CsvLineDecoder.parse("");
+        assertEquals(0, r.size());
```
jowerner, line 62:
```suggestion
        test("", "");
```
I would expect the decoder to return a list with a single empty field in this case. This would be in line with the justDelimiters() test farther down and also the good old CsvUtils class.

src/test/java/com/xceptance/common/util/CsvLineDecoderTest.java
```diff
@@ -0,0 +1,374 @@
...
+    {
+        testException("'''", "Quoted field with quotes was not ended properly at: 3");
+        testException("'''''", "Quoted field with quotes was not ended properly at: 5");
+        testException("'','''", "Quoted field with quotes was not ended properly at: 6");
+        testException("'','''',''',''", "Quoted field with quotes was not ended properly at: 14");
+    }
+
+    /*
+     * Extra concept tests for bit manipulation
+     */
```
jowerner, line 260: Do we really need these bit tests?

jowerner (changes requested):
See comments regarding CsvLineDecoderTest.

I would also prefer to leave the *Data classes as they are. Moving the CSV decoding part out of these classes can be done in XLT 8.

src/test/java/com/xceptance/common/util/CsvLineDecoderTest.java
rschwietzke: Removed

src/test/java/com/xceptance/common/util/CsvLineDecoderTest.java
rschwietzke: Done.

PR #422, opened 2023-10-05
421 mergerule processing split

rschwietzke:
* fixed missing hashCode caching in XltCharBuffer
* removed single instance concept of request merge rule processors, each thread has its own merge processor and its own cache
* FastHashMap compares the hashcode first before it equals the key to safe some recurring operations

We gain about 1 min for a test report that was 5 min before, it is 4 min now.


src/main/java/com/xceptance/common/collection/FastHashMap.java
jowerner, line 1: The methods `put` and `remove` might benefit from a similar improvement.

src/main/java/com/xceptance/xlt/report/DataParserThread.java
```diff
@@ -345,6 +347,8 @@ else if (state == ReturnState.STOP)
                 final String msg = String.format("Failed to apply request merge rule: %s\n%s", requestProcessingRule, t);
                 LOG.error(msg);
 
+                t.printStackTrace();
+
```
jowerner, line 351: Please remove.

src/main/java/com/xceptance/xlt/report/mergerules/AbstractPatternRequestFilter.java
```diff
@@ -32,26 +32,14 @@
...
-     * The size of the cache, because different filters have different demands
-     */
-    private final int cacheSize;
+    private final LRUFastHashMap<CharSequence, Matcher> cache;
 
     /**
      * Just a place holder for a NULL
      */
-    private static final Matcher NULL = Pattern.compile(".*").matcher("null");
+    public static final Matcher NULL = Pattern.compile(".*").matcher("null");
```
jowerner, line 42: Why public now?

src/main/java/com/xceptance/xlt/report/mergerules/UrlRequestFilter.java
```diff
@@ -45,7 +45,7 @@ public UrlRequestFilter(final String regex, final boolean exclude)
     {
         // we don't want to cache here due to the large variance in
         // urls, it is too costly to look things up with a lot of cache misses
-        super("u", regex, exclude, 0);
+        super("u", regex, exclude, 1000);
```
jowerner, line 48: The cache is in use now. Is this by accident? Anyway, we need to either adjust the comment or set the cache size back to 0.

rschwietzke: It is not longer expensive to look that up and we also cache non-hits. Seems to work. Larger cache does not show an effect.

src/main/java/com/xceptance/xlt/report/mergerules/AbstractPatternRequestFilter.java
rschwietzke: It was supposed to be used outside that that did not work. Changed.

src/main/java/com/xceptance/common/collection/FastHashMap.java
rschwietzke: Will make it another PR later and measure that first. We do more get than put.

src/test/java/com/xceptance/xlt/api/util/XltCharBufferTest.java
```diff
@@ -1088,6 +1122,105 @@ public void equalsTest()
...
+
+        // identical by hashcode, hashcode touched
+        {
+            var s1 = XltCharBuffer.valueOf("foo");
+            var s2 = XltCharBuffer.valueOf("foo");
+            // caches hash codes
+            assertEquals(s1.hashCode(), s2.hashCode());
+            assertTrue(s1.equals(s2));
+            assertTrue(s2.equals(s1));
+        }
```
jowerner, line 1134: Seems to be the same as the following block. Remove?

src/test/java/com/xceptance/xlt/api/util/XltCharBufferTest.java
```diff
@@ -1088,6 +1122,105 @@ public void equalsTest()
...
+        // one has cached hashcode, was has not
+        {
+            var s1 = new XltCharBufferSameHash("foo2".toCharArray(), 1);
+            var s2 = new XltCharBufferSameHash("foo3".toCharArray(), 0);
+            // just check that the override is right
+            assertEquals(1, s1.hashCode());
+            assertEquals(0, s2.hashCode());
+            assertFalse(s1.equals(s2));
+            assertFalse(s2.equals(s1));
+        }
```
jowerner, line 1206: All these tests mess with the hash code a lot, but the `equals` method is actually completely unaffected by the hash code. Remove?

jowerner (changes requested):
See comments inline.

src/test/java/com/xceptance/xlt/api/util/XltCharBufferTest.java
rschwietzke: Removed

src/test/java/com/xceptance/xlt/api/util/XltCharBufferTest.java
rschwietzke: Removed

jowerner (approved):

PR #424, opened 2023-10-06
423 disable js engine

rschwietzke:
JS engine is off by default and not started. A short test has shown that the scripts for regular load tests still work. We are also not creating the debugger object, if not needed.

## CPU
-10 %

### Old
image

### New
image

## Memory
-200 MB

### Old
image

### New
image

### DIFF
--- a/samples/testsuite-template/config/default.properties
+++ b/samples/testsuite-template/config/default.properties
@@ -182,18 +182,18 @@ com.xceptance.xlt.staticContentCache.size = 1000
 ################################################################################
 
 ## Whether the JavaScript Engine is enabled. Default is false.
-## The JS engine of the HtmlUnit webclient is heavy, if we don't need
-## it, we don't start it
-# com.xceptance.xlt.javaScriptEngineEnabled = false
+## The JS engine of the HtmlUnit web client is heavy so if you don't need it,
+## don't enable it.
+#com.xceptance.xlt.javaScriptEngineEnabled = false
 
-## Whether JavaScript is enabled AND the JS engine has been started too.
-## Setting this to false still starts the JS engine (if set top true) but does not
-## execute any JS in the page. Default is false.
-# com.xceptance.xlt.javaScriptEnabled = false
+## Whether the execution of JavaScript code in a Web page is enabled. Default
+## is false. When enabling it, ensure the javaScriptEngineEnabled flag above
+## is true as well.
+#com.xceptance.xlt.javaScriptEnabled = false
 
 ## Whether the JavaScript debugger (function call logger) is enabled.
 ## Default is false.
-com.xceptance.xlt.js.debugger.enabled = false
+#com.xceptance.xlt.js.debugger.enabled = false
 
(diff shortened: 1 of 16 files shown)


rbri:
cool

samples/testsuite-template/config/default.properties
h-arlt, line 1: Apply changes to the default configuration in the other sample projects too?

h-arlt (changes requested):

samples/testsuite-template/config/default.properties
rschwietzke: That would be perfect but because we go with off as default, it is less pressing to update the others. They get it indirectly.

rschwietzke: For the XLT sample projects, we can apply that change now, for the external projects, we apply it as part of the update to 7.whatever

jowerner (changes requested):
XltWebClient always sets an XltJavaScriptEngine instance at the moment. To save even more resources, this should be skipped if `javaScriptEngineEnabled` is false.

When loading static content, the XltWebClient relies on the Javascript engine being available in order to evaluate CSS styles, etc. This functionality of XltWebClient needs to be bypassed when javaScriptEngineEnabled is false.

Also it seems that we should rather replace most occurrences of `WebClient.getOptions().isJavaScriptEnabled()` with `WebClient.isJavaScriptEnabled()` as the latter also takes the `javaScriptEngineEnabled` flag into account.

rbri:
>When loading static content, the XltWebClient relies on the Javascript engine being available in order to evaluate CSS styles, etc. 

I guess this is now fixed. Do you have some time to test this?

rbri:
just made a new snapshot build...

src/main/java/com/xceptance/xlt/engine/XltWebClient.java
```diff
@@ -1574,15 +1577,18 @@ private List<CSSStyleRuleImpl> getCssRulesWithUrls(final HtmlPage page)
     {
         final List<CSSStyleRuleImpl> cssRules = new ArrayList<CSSStyleRuleImpl>();
 
-        final Window window = ((Window) page.getEnclosingWindow().getScriptableObject());
-        final HTMLDocument document = ((HTMLDocument) window.getDocument());
-
-        // check all style sheets
-        final StyleSheetList sheets = document.getStyleSheets();
-        for (int i = 0; i < sheets.getLength(); i++)
+        if (isJavaScriptEnabled())
```
rbri, line 1584: with 3.7 js support is no longer required to work with style sheets.

At least for this you can hopefully use HtmlPage.getStyleSheets() https://github.com/HtmlUnit/htmlunit/commit/03877bb3ba4e163523198192dd96ab7b37c0c7f2.
Will try to make a new snapshot later today.

jowerner: Cool! When time permits, we will upgrade to 3.7 and see what we can simplify on our end.

jowerner (approved):

Issue #179, opened 2022-01-03
Support for JDK 17+


js-xc:
With older Java versions, running XLT test scenarios (in the IDE) caused the following console warning: 
`WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.xceptance.xlt.engine.socket.InstrumentedSocketImpl (file:/home/julia/.m2/repository/com/xceptance/xlt/5.7.0/xlt-5.7.0.jar) to constructor java.net.PlainSocketImpl()
WARNING: Please consider reporting this to the maintainers of com.xceptance.xlt.engine.socket.InstrumentedSocketImpl
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release`

This "future release" seems to have been reached with JDK 17, when the warning turns into an error and running tests becomes impossible: 
`[11:43:11,425] ERROR [TBrowse-0] - Execution step failed (user: 'TBrowse-0', output: '1640083389589'): null
java.lang.ExceptionInInitializerError
    at com.xceptance.xlt.engine.socket.InstrumentedSocketImplFactory.createSocketImpl(InstrumentedSocketImplFactory.java:33)
    at java.base/java.net.Socket.setImpl(Socket.java:548)
    at java.base/java.net.Socket.<init>(Socket.java:143)
    at java.base/javax.net.DefaultSocketFactory.createSocket(SocketFactory.java:265)
    at org.apache.http.conn.ssl.SSLConnectionSocketFactory.createSocket(SSLConnectionSocketFactory.java:344)
    at org.apache.http.impl.conn.DefaultHttpClientConnectionOperator.connect(DefaultHttpClientConnectionOperator.java:119)
    at org.apache.http.impl.conn.PoolingHttpClientConnectionManager.connect(PoolingHttpClientConnectionManager.java:376)
    at org.apache.http.impl.execchain.MainClientExec.establishRoute(MainClientExec.java:393)
    at org.apache.http.impl.execchain.MainClientExec.execute(MainClientExec.java:236)
    at org.apache.http.impl.execchain.ProtocolExec.execute(ProtocolExec.java:186)
    at org.apache.http.impl.execchain.RetryExec.execute(RetryExec.java:89)
    at org.apache.http.impl.execchain.RedirectExec.execute(RedirectExec.java:110)
    at org.apache.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:185)
[...]

PR #435, opened 2023-10-24
#179: Support for JDK 17+

jowerner:
XLT does no longer throw if the socket layer cannot be instrumented with the used JVM, but logs a warning only.

### DIFF
--- a/src/main/java/com/xceptance/xlt/engine/socket/XltSockets.java
+++ b/src/main/java/com/xceptance/xlt/engine/socket/XltSockets.java
@@ -15,9 +15,12 @@
  */
 package com.xceptance.xlt.engine.socket;
 
-import java.io.IOException;
 import java.net.Socket;
 import java.net.SocketImplFactory;
+
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


src/main/java/com/xceptance/xlt/engine/socket/XltSockets.java
```diff
@@ -46,9 +51,9 @@ public final class XltSockets
                 // set the global socket impl factory
                 Socket.setSocketImplFactory(new InstrumentedSocketImplFactory());
             }
-            catch (final IOException ex)
+            catch (final Throwable ex)
             {
-                throw new RuntimeException("Failed to initialize XLT sockets", ex);
+                LOG.warn("Failed to initialize XLT sockets: {}", ExceptionUtils.getRootCauseMessage(ex));
```
h-arlt, line 56: Logging a warning might be fine for development mode but when running in context of a load test, we may log it as error (or even re-throw it) instead. WDYT? Otherwise, load test environment misconfigurations might be easily overseen.

h-arlt: Nice. Thx.

h-arlt (approved):

Issue #390, opened 2023-07-18
Issues with the new timer recorder plugin for Chrome


jowerner:
There have been reports that test scenarios using `XltChromeDriver` show different issues now, such as:

* higher scenario runtimes
* higher request runtimes
* unexpected "think times" at the end of a scenario
* unexpected errors when quitting the driver (`UnhandledAlertException: unexpected alert open: {Alert text : }`) 

Looks as if this is related to our timer recorder plugin for Chrome. In XLT 7.0.0, we have updated it to support Chrome Manifest v3 (see #141). A cross-check with XLT 6.x showed that the new plugin causes these issues with both XLT 6 and 7, while the old plugin does not make any trouble.

PR #440, opened 2023-11-20
#390: Issues with the new timer recorder plugin for Chrome

jowerner:
Chrome Timer Recorder
* brought Manifest v3 code back
* upped minimum chrome version to 116 (see https://developer.chrome.com/docs/extensions/mv3/service_workers/service-worker-lifecycle/#chrome-116)
* installed a periodic task that sends a dummy message to XLT to keep the service worker alive
* made the usage of the session storage configurable (off by default)
* added a helper function to calculate the size of the request line
* minor optimizations
* added some useful logger functions

XLT
* made the usage of the session storage in the chrome timer recorder configurable (off by default)
* handle new keep-alive messages by ignoring them
* adjusted logging when receiving timer data

### DIFF
--- a/src/main/java/com/xceptance/xlt/clientperformance/WebExtConnectionHandler.java
+++ b/src/main/java/com/xceptance/xlt/clientperformance/WebExtConnectionHandler.java
@@ -168,16 +168,24 @@ public class WebExtConnectionHandler implements ConnectionListener
     @Override
     public void onMessage(ClientPerformanceExtensionConnection connection, JSONObject data, Responder responder)
     {
+        final String action = data.optString("action");
+
+        LOG.debug("Message received: {}", action);
+
         if (LOG.isTraceEnabled())
         {
-            LOG.trace("Message received: " + data);
+            LOG.trace("Message data: " + data);
         }
 
         try
         {
-            if ("DUMP_PERFORMANCE_DATA".equals(data.optString("action")))
+            if ("DUMP_PERFORMANCE_DATA".equals(action))
             {
                 dumpPerformanceData(data.optString("performanceData"));
+            }
+            else if ("KEEP_ALIVE_PING".equals(action))
+            {
+                // ignore for now
             }
         }
         catch (Throwable t)
@@ -213,11 +221,25 @@ public class WebExtConnectionHandler implements ConnectionListener
         {
             final List<ClientPerformanceData> performanceData = PerformanceDataTransformator.getTransformedPerformanceDataList(rawData);
 
-            ClientPerformanceMetrics.updatePerformanceData(session, performanceData);
-
             if (LOG.isDebugEnabled())
             {
-                LOG.debug("Dumped client-performance metrics: " + rawData);
+                int requestTimings = 0;
+                int pageLoadTimings = 0;
+
+                for (final ClientPerformanceData clientPerformanceData : performanceData)
+                {
+                    requestTimings += clientPerformanceData.getRequestList().size();
+                    pageLoadTimings += clientPerformanceData.getCustomDataList().size();
+                }
+
+                LOG.debug("Received client-performance metrics (requests: {}, page load timings: {})", requestTimings, pageLoadTimings);
+            }
+
+            ClientPerformanceMetrics.updatePerformanceData(session, performanceData);
+
+            if (LOG.isTraceEnabled())
+            {
+                LOG.trace("Dumped client-performance metrics: " + rawData);
             }
         }
         catch (final Throwable t)
@@ -229,7 +251,7 @@ public class WebExtConnectionHandler implements ConnectionListener
     /**
      * Fetch the remaining timing data from the client which were not sent.
      */
-    public String fetchPerformanceData() throws JSONException, TimeoutException, CommunicationException, InterruptedException
+    private String fetchPerformanceData() throws JSONException, TimeoutException, CommunicationException, InterruptedException
     {
         ClientPerformanceExtensionConnection conn = currentConnection;
         if (conn == null || !conn.isOpen())
--- a/src/main/java/com/xceptance/xlt/clientperformance/WebExtConnectionHandler.java
+++ b/src/main/java/com/xceptance/xlt/clientperformance/WebExtConnectionHandler.java
@@ -170,12 +170,7 @@ public class WebExtConnectionHandler implements ConnectionListener
     {
         final String action = data.optString("action");
 
-        LOG.debug("Message received: {}", action);
-
-        if (LOG.isTraceEnabled())
-        {
-            LOG.trace("Message data: " + data);
-        }
+        LOG.debug("Message received for action: {}", action);
 
         try
         {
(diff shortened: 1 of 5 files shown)


jowerner:
Some of the changes needs to be made for the FF timer recorder as well. Will do this when the review for the Chrome version is done.

src/main/java/com/xceptance/xlt/clientperformance/WebExtConnectionHandler.java
```diff
@@ -168,17 +168,25 @@ public void onConnect(ClientPerformanceExtensionConnector connector, ClientPerfo
...
     public void onMessage(ClientPerformanceExtensionConnection connection, JSONObject data, Responder responder)
     {
+        final String action = data.optString("action");
+
+        LOG.debug("Message received: {}", action);
+
         if (LOG.isTraceEnabled())
         {
-            LOG.trace("Message received: " + data);
+            LOG.trace("Message data: " + data);
```
h-arlt, line 177: One of the log statements seems superfluous to me. It should be sufficient to log the entire message data and remove the first log statement that logs the value of the action property only. WDYT?

xlt-timerrecorder-chrome/src/background.js
```diff
@@ -234,281 +294,351 @@ function sendTimingDataEntries(tabId, onlyFinished) {
...
+    if (hasTimingDataEntry(TimingData, tabId)) {
+      setTimingData(getCurrentTimingDataEntry(TimingData, tabId), message.value);
     }
   }
+
+  storeStorageCache();
+});
+
+chrome.runtime.onSuspend.addListener(function () {
+  storeStorageCache();
```
h-arlt, line 617: Now that use of session storage is optional, what about to log an error here in case service worker is suspended but use of session storage is disabled? Doing so might help to figure out whether period pings prevent the service worker to be suspended or not.

xlt-timerrecorder-chrome/src/background.js
```diff
@@ -1,21 +1,60 @@
+// log something so we can see when a service worker was (re-)activated
+log("Service worker activated");
+
+// send a keep-alive message every 10 secs to prevent the service worker from being deactivated after 30 secs of inactivity
+setInterval(
+  () => {
+    log("Sending keep-alive message");
+    send({ action: "KEEP_ALIVE_PING", data: "[]" }, "<dummy>");
```
h-arlt, line 8: We might adjust function `send` to simply have `messageID` undefined when not given. This would eliminate the neef for the `<dummy>` argument here.

```
function send(data, messageID) {
  const messageString = JSON.stringify({
    messageID: messageID !== undefined ? String(messageID) : undefined,
    data : data
  });
  ... 
}
```

src/main/java/com/xceptance/xlt/clientperformance/WebExtConnectionHandler.java
jowerner: The message with all the request and event details might get big, for example half an MB. Printing this amount of data as a one-liner to the console slows down my Eclipse considerably and scrolling the console gets sluggish.

For this reason, I would like to log only the type of message just arrived and maybe how many entries I received if the message was a details message. If needed, the logging of the raw data can still be enabled.

xlt-timerrecorder-chrome/src/background.js
jowerner: Tests have shown that the `onSuspend` event is not fired when the service worker is deactivated. To my knowledge, there is also no other event for that. Looks like they don't want extensions to delay their deactivation in any way.

xlt-timerrecorder-chrome/src/background.js
jowerner: I wonder if we need the parameter `messageID` at all. The extension either leaves it undefined or sets only what XLT had passed, however, XLT does not send one either. WDYT about removing this?

xlt-timerrecorder-chrome/src/background.js
h-arlt: AFAIK this event was introduced to let service workers be notified in case they will be suspended. Thus, it rather seems to be a bug in case this event is never fired.
Anyway, it should not hurt to have an additional log statement here as well; just in case Chrome developers decide to fix the missing event dispatch.

src/main/java/com/xceptance/xlt/clientperformance/WebExtConnectionHandler.java
h-arlt: Well, we might think about to limit the output by adjusting `Message#toString()` (e.g. `StringUtils.abbreviate`) in any case.

We may keep the 1st log statement, but its message is little bit confusing as `action` is not the message received; `Message received for action '{}'` fits better IMO.

I also noticed that `ClientPerformanceExtensionConnector` already logs the deserliazed message at level TRACE. So, the entire if-statement is indeed superfluous.

xlt-timerrecorder-chrome/src/background.js
jowerner: Indeed, good you've found that! So I better do not change anything here. Will apply your suggestion instead.

src/main/java/com/xceptance/xlt/clientperformance/WebExtConnectionHandler.java
jowerner: Since the whole message is indeed printed elsewhere, I will remove the logger call here and another one farther below.

xlt-timerrecorder-chrome/src/background.js
jowerner: In fact, XLT requires `messageID` to exist in the JSON message, so we have to set something, even if it is the string "undefined".

h-arlt: Well... so be it ;)

jowerner:
Marked this PR as draft so it won't be merged accidentally before these steps are done:
* Some of the changes need to be done also for the Firefox timer recorder after review for Chrome is complete.
* Chrome extensions cannot be packaged right now because of a bug in Chrome 119. Once this is fixed, we need to undo the temporary zip-packaging.

xlt-timerrecorder-chrome/src/background.js
```diff
@@ -1,21 +1,60 @@
+// log something so we can see when a service worker was (re-)activated
+log("Service worker activated");
+
+// send a keep-alive message every 10 secs to prevent the service worker from being deactivated after 30 secs of inactivity
+setInterval(
+  () => {
```
h-arlt, line 6: Just for consistency reasons. We may rework entire code base to use modern ECMAScript features such as arrow functions, async/await, destructuring assignment etc. later.
 
```suggestion
  function ping() {
```

h-arlt (approved):

jowerner:
> Marked this PR as draft so it won't be merged accidentally before these steps are done:
> 
>     * Some of the changes need to be done also for the Firefox timer recorder after review for Chrome is complete.
> 
>     * Chrome extensions cannot be packaged right now because of a bug in Chrome 119. Once this is fixed, we need to undo the temporary zip-packaging.

To not delay the fixes for the Chrome timer recorder any further, we postpone the first point to #462. The second point is done.

h-arlt (approved):

Issue #448, opened 2023-11-29
Report generator: Link to result browser dirs only if there is an index.html


jowerner:
When generating the *Errors* page, the report generator may add a result browser link to an error table entry. It does this if it was able to find a corresponding result browser directory.

However, sometimes test suites store additional data files in the result browser directory for later analysis. This causes the report generator to link to this directory, even if this directory does not contain result browser files (due to the error limiter). Clicking on this link would then result in a Not Found page because there is no `index.html` file. Only if such an `index.html` file exists, should the report generator link to this directory.

PR #449, opened 2023-11-29
#448:  Report generator: Link to result browser dirs only if there is an index.html

jowerner:


src/main/java/com/xceptance/xlt/report/providers/ErrorsReportProvider.java
```diff
@@ -465,42 +465,35 @@ public void processDataRecord(final Data stat)
                     final String directoryHint = txnStats.getDumpDirectoryPath();
                     if (directoryHint != null)
                     {
+                        // check if we should add/replace a result browser directory hint
                         final int size = errorReport.directoryHints.size();
-                        if (size < directoryLimitPerError)
+                        if (size < directoryLimitPerError || random.nextDoubleFast() <= directoryReplacementChance)
```
h-arlt, line 470: Small optimization: extract the expression `size < directoryLimitPerError` into a local variable, e.g. `safeToAdd`, and re-use it here and at line 481.

h-arlt (approved):

src/main/java/com/xceptance/xlt/report/providers/ErrorsReportProvider.java
jowerner: Done.

PR #454, opened 2023-12-14
451 footer information

js-xc:


jowerner:
In `config/xsl/loadreport/index.xsl`, please also replace 
```
 <xsl:with-param name="projectName" select="configuration/projectName" />
```
with
```
 <xsl:with-param name="projectName" select="$projectName" />
```
for consistency when calling the `head` template.

jowerner:
Now that we started it, we should rework the `header` template in the same way, i.e. remove default parameter values and pass parameters explicitly in the load test report templates. WDYT?

jowerner (changes requested):
See comments.

js-xc:
Congratulations, you found the one file where I forgot to update the param for the `head` template :joy:

jowerner (approved):

PR #455, opened 2024-01-11
429 additional report metrics

js-xc:
Not sure what will happen if input reports are not using the same Pxx (usually 50/95/99/99.9), I will check this.

### DIFF
--- a/config/xsl/diffreport/util/timer-table.xsl
+++ b/config/xsl/diffreport/util/timer-table.xsl
@@ -29,7 +29,7 @@
                         <xsl:if test="$type = 'transaction'">
                             <th class="colgroup1">Events</th>
                         </xsl:if>
-                        <th colspan="5">Runtime [ms]</th>
+                        <th colspan="4">Runtime [ms]</th>
                         <xsl:if test="$percentileCount &gt; 0">
 	                        <th colspan="{$percentileCount}" class="colgroup1">Runtime Percentiles [ms]</th>
 	                    </xsl:if>
@@ -46,7 +46,6 @@
                         <xsl:if test="$type = 'transaction'">
                             <th class="table-sortable:numeric colgroup1">Total</th>
                         </xsl:if>
-                        <th class="table-sortable:numeric" title="The median of the data series.">Med.</th>
                         <th class="table-sortable:numeric" title="The arithmetic mean of the data series.">Mean</th>
                         <th class="table-sortable:numeric" title="The smallest value of the data series.">Min.</th>
                         <th class="table-sortable:numeric" title="The largest value of the data series.">Max.</th>
--- a/config/xsl/diffreport/util/timer-table.xsl
+++ b/config/xsl/diffreport/util/timer-table.xsl
@@ -85,19 +85,39 @@
 	                                <xsl:value-of select="11 + $percentileCount"/>
 	                            </xsl:when>
 	                            <xsl:when test="$type = 'transaction'">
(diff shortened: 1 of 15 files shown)


jowerner:
Now that we have the Percentiles column set in the table, we can remove the Median column in the Runtime column set.

jowerner:
In case there is no data available to show in a table, the then empty table misses some cells in the table body and footer.

jowerner:
Add an asterisk to the 1/h and 1/d column headers in the Bandwidth table to indicate these values are projected.

js-xc:
> Add an asterisk to the 1/h and 1/d column headers in the Bandwidth table to indicate these values are projected.

This applies to diffreport as well as loadtest report, right?

jowerner:
> > Add an asterisk to the 1/h and 1/d column headers in the Bandwidth table to indicate these values are projected.
> 
> This applies to diffreport as well as loadtest report, right?

Right.

jowerner:
While we are at it, could you please add an explanation for the two asterisks at the Distinct** column to the help text of the Requests page? Have a look at the load test report and borrow the text from there.

jowerner:
In `config/xsl/loadreport/util/timer-row.xsl`, we show the "Distinct" column for requests only. Please simplify the template by using a plain `xsl:if` instead of `xsl:choose`.

js-xc:
> In case there is no data available to show in a table, the then empty table misses some cells in the table body and footer.

I don't understand, what's wrong with this? Or is the html wrong and Firefox is just filling the gaps itself so that it looks ok?
Screenshot from 2024-01-15 13-50-10

jowerner:
> > In case there is no data available to show in a table, the then empty table misses some cells in the table body and footer.
> 
> I don't understand, what's wrong with this? Or is the html wrong and Firefox is just filling the gaps itself so that it looks ok? Screenshot from 2024-01-15 13-50-10

See below:

Screenshot 2024-01-15 at 13-53-58 XLT Performance Comparison Report - Page Load Timings

config/xsl/diffreport/util/timer-table.xsl
```diff
@@ -0,0 +1,116 @@
...
+	                                <xsl:value-of select="12 + $percentileCount"/>
+	                            </xsl:when>
+	                            <xsl:when test="$type = 'action'">
+	                                <xsl:value-of select="12 + $percentileCount"/>
+	                            </xsl:when>
+	                            <xsl:otherwise>
+	                                <xsl:value-of select="11 + $percentileCount"/>
+	                            </xsl:otherwise>
+	                        </xsl:choose>
+                        </xsl:variable>
```
jowerner, line 97: The calculated columns value looks incorrect to me, except for requests. But see the other comment below as well.

config/xsl/diffreport/util/timer-table.xsl
```diff
@@ -0,0 +1,116 @@
...
+                        <tfoot>
+                            <tr>
+                                <td colspan="{$columns}"></td>
+                            </tr>
+                            <xsl:call-template name="filtered-footer-row" />
+                        </tfoot>
```
jowerner, line 103: When using a single td with colspan, the column (set) striping effect as introduced in the table header is lost. We should have at least one td per column (set). Make sure to match the look with the Bandwidth table.

(I know that column striping in general is a mess in the diff report, but definitely another topic. Just wanted to make sure that we don't add to it.)

jowerner (approved):

PR #468, opened 2024-02-13
consider action name on error dump count

js-xc:


src/main/java/com/xceptance/xlt/engine/resultbrowser/ErrorCounter.java
```diff
@@ -201,7 +201,7 @@ public boolean countDumpIfOpen(final SessionImpl session)
         if (session.hasFailed())
         {
             // get the error key
-            final String key = getErrorKey(session.getUserName(), session.getFailReason());
+            final String key = getErrorKey(session.getUserName()+session.getCurrentActionName(), session.getFailReason());
```
jowerner, line 204: This did not work in my mini load test, because the current action name was always null. I had to change the code to use `session.getFailedActionName()` to make it work. In addition, I would like to separate the two parts by a "|" to reduce the chance for ambiguities.

```suggestion
            final String key = getErrorKey(session.getUserName() + "|" + session.getFailedActionName(), session.getFailReason());
```

jowerner (changes requested):
See comments.

src/main/java/com/xceptance/xlt/engine/resultbrowser/ErrorCounter.java
jowerner: An even better solution would be to pass the failed action name as additional argument to `getErrorKey()`. This involves adjusting some test cases, though.

jowerner (approved):

PR #475, opened 2024-02-23
crosshair on charts

js-xc:
(very basic functionality)


rschwietzke (changes requested):
I like it. Please update the third-party license section of XLT regarding the lib and its license.

rschwietzke (approved):
Thanks.

PR #479, opened 2024-03-11
474 error percentages

js-xc:
percentage calculation now in java code instead of xsl


js-xc:
some of the agent summary values in xml are still not correct (but aren't used anywhere either), I guess that's due to the AgentDataProcessor just updating values for a timestamp it already knows (i.e. when 3 agents all recorded data at the same second), but I have no idea how we'd change that without rewriting the whole thing...

src/main/java/com/xceptance/xlt/report/providers/RuntimeIntervalData.java
```diff
@@ -0,0 +1,40 @@
...
+{
+    /**
+     * The number of timer values per configured runtime interval.
+     */
+    public int count;
+
+    /**
+     * The percentage of timer values per configured runtime interval.
+     */
+    public BigDecimal percent;
```
jowerner, line 33: Please rename to `percentage`.

src/main/java/com/xceptance/xlt/report/providers/AgentReport.java
```diff
@@ -45,4 +45,6 @@ public class AgentReport
     public int transactions;
 
     public int transactionErrors;
+
+    public BigDecimal transactionErrorsPercent;
```
jowerner, line 49: Please rename to `transactionErrorPercentage`.

jowerner:
The percentage values for the individual request runtime segments as shown in the table **footer row** are currently all "NaN%". Maybe you need to adjust the footer row template as well?

src/main/java/com/xceptance/xlt/report/providers/BasicTimerDataProcessor.java
```diff
@@ -121,6 +121,11 @@ public TimerReport createTimerReport(final boolean generateHistogram)
         // set the counts
         final double count = runTimeStatistics.getCount();
         final long duration = Math.max((getEndTime() - getStartTime()) / 1000, 1);
+        
+        if (count != 0)
+        {
+            timerReport.errorPercentage = ReportUtils.convertToBigDecimal(100 * totalErrors / count);
+        }
```
jowerner, line 128: Elsewhere you set the percentage to 0 in case the count is 0. Do the same here as well?

js-xc: Does not really matter because it is implicitly set to 0 if it was never specified. I can change it if you think it's better (readable?) that way.

jowerner: Please set it to 0. `errorPercentage` is a BigDecimal and would otherwise either be null or missing in the XML, not sure.

src/main/java/com/xceptance/xlt/report/providers/BasicTimerDataProcessor.java
```diff
@@ -121,6 +121,15 @@ public TimerReport createTimerReport(final boolean generateHistogram)
...
+        else
+        {
+            timerReport.errorPercentage = ReportUtils.convertToBigDecimal(0);
+        }
```
jowerner, line 132: The percentage calculation is done using integer arithmetic, so the result may be wrong. 

How about creating a helper method like `BigDecimal ReportUtils.calculatePercentage(int count, int totalCount)` that takes care of this subtlety and also includes the check if count is 0? Use this method wherever we calculate error percentages.

src/main/java/com/xceptance/xlt/report/util/ReportUtils.java
```diff
@@ -338,4 +340,24 @@ public static String obtainProjectName(final Iterable<Document> documents)
...
+     * Calculates the percentage given by count/total, as a {@link BigDecimal} (rounding it to have at most
+     * {@value #DEFAULT_DECIMAL_PLACES} decimal places). Returns 0 if total is 0.
+     * 
+     * @param count
+     * @param total
+     * @return
+     */
+    public static BigDecimal calculatePercentage(int count, int total)
+    {
+        if (count == 0)
```
jowerner, line 354: Should be `total` instead of `count`, right?

js-xc: :facepalm:

src/main/java/com/xceptance/xlt/report/providers/AgentReport.java
jowerner: Almost there. Mind the singular form in my suggestion.

js-xc:
This is waiting for @jowerner to decide whether we'd want to rename other pieces of the report as well, right? Just a thought: the more we restructure/rename the report, the more the comparison report will struggle between old and new data (like, "let's compare to last year's test run"), might this be an issue?

jowerner (approved):

PR #483, opened 2024-03-19
480 added web vitals to comparison report

js-xc:


jowerner:
The same applies to all current Web Vitals: the smaller the values, the better. In this context, please check the red/green colorization again. 

* FCP, LCP, INP, TTFB: Reverse red/green coloring.
* CLS, FID: Currently showing red for both increments and decrements?

jowerner:
From the description, remove the information about the good/improve/poor coloring of score values.

jowerner:
If we have a "no data" cell, i.e. no score value for a certain web vital in both input reports, we still show a tooltip to indicate the "change". What do you think about not showing the tooltip in these cases?

PR #490, opened 2024-04-15
489 diff custom values

js-xc:
I tried it with "neutral" coloring and like it better than no coloring at all. But just my opinion, please check it, if you like it better without highlighting we can just roll back the last two commits.


config/testreport/css/default.css
```diff
@@ -1599,7 +1612,19 @@ tr:hover td.n60 , tr:hover td.n61 , tr:hover td.n62 , tr:hover td.n63 , tr:hover
...
+tr:hover td.a80, tr:hover td.a81, tr:hover td.a82, tr:hover td.a83, tr:hover td.a84, tr:hover td.a85, tr:hover td.a86, tr:hover td.a87, tr:hover td.a88, tr:hover td.a89 { background: linear-gradient(rgba(107,139,174,0.7), rgba(0,0,0,0), rgba(107,139,174,0.7)) !important; }
+tr:hover td.a90, tr:hover td.a91, tr:hover td.a92, tr:hover td.a93, tr:hover td.a94, tr:hover td.a95, tr:hover td.a96, tr:hover td.a97, tr:hover td.a98, tr:hover td.a99 { background: linear-gradient(rgba(91,126,164,0.7), rgba(0,0,0,0), rgba(91,126,164,0.7)) !important; }
+tr:hover td.a100, tr:hover .infinity.neutral                                   																						     { background: linear-gradient(rgba(81,114,148,0.7), rgba(0,0,0,0), rgba(81,114,148,0.7)) !important; }                                                                                                                                                              { background: linear-gradient(rgb(200,120,120), rgba(0,0,0,0), rgb(200,120,120)) !important; }
```
jowerner, line 1627: Looks like the CSS is somewhat damaged. The `{ background: linear-gradient(rgb(200,120,120), rgba(0,0,0,0), rgb(200,120,120)) !important; }` part at the end of the line belongs to the `tr:hover td.n100` line above.

jowerner (approved):

PR #500, opened 2024-05-23
465 responsive result browser

js-xc:
First version, let's start with this and add more when needed. 
So far: 
* bigger font in menu
* column layout in mobile/portrait mode
* filter menu made usable on mobile (full screen overlay)
* navigation can be expanded/hidden on mobile in portrait mode
* realbrowser action screenshots are sized to 100% screen width per default, zoomable

### DIFF
--- a/resultbrowser/src/css/default.css
+++ b/resultbrowser/src/css/default.css
@@ -659,7 +659,6 @@ table.key-value-table td.value .csep {
 	font-size: 20px;
 }
 
-
 /********************************************************************************************
  *
  * Responsive
@@ -672,6 +671,10 @@ table.key-value-table td.value .csep {
 }
 
 #header .expander {
+	display: none;
+}
+
+#menu-close {
 	display: none;
 }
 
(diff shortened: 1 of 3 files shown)


rschwietzke (approved):
Verified the demo. Approved.

resultbrowser/src/css/default.css
```diff
@@ -627,3 +632,242 @@ table.key-value-table td.value .csep {
...
+	flex-flow: row;
+	width: 100vw !important;
+}
+
+#header .expander {
+	display: none;
+}
+
+#menu-close {
+	display: none;
```
jowerner, line 678: How about adding `cursor: pointer;`?

js-xc: To which element? `#menu-close` is only ever visible in mobile mode, where you don't see a cursor.

jowerner: Haha, true!

Issue #478, opened 2024-02-28
Labeling of Transactions, Actions, and Requests


rschwietzke:
# Goal
We want to label transactions, actions, and requests. So we can group these better. See https://training.xceptance.com/qa-and-test/050-performance-rating-system-devtest-leeds-2019.html#/14 for an example of categories (homepage, catalog, cart, account, checkout, place order).

* These labels have to be part of the testreport XML and later HTML.
* For performance reasons, this labeling has to happen when all data was collected because it does not merge or split data, just "stamps" it. 
* A label can be any string, such as homepage, #test, F17, Foo Bar or even an emulated double label, such as "homepage, B2C" (w/o the quotes of course). Barely any limits on what characters are possible.

## Examples

* Transactions: transactional, registered
* Actions: catalog, cart, checkout, homepage
* Requests: catalog, cart, checkout, homepage

# Non-Goal
We would rather not add multiple labels at the moment, so we don't think tags here.

# Requirements
## Report
All label information has to be part of the output XML so the reporting can later on incorporate it as well as the upcoming evaluation module.

The transaction, action, and the request table in the section of the report will get a new column "Label" (next to name) which supports sorting and filtering similarly to "Name". One should be able to filter one column and later the other. We imagine two input fields here so it is clear where you filter, and we don't introduce any special syntax for the filters by that.

When there is no label attached, the cell or column stays empty. 

## Definition
The test suite gets new properties for the labeling. We will use a similar concept to merge and Apdex rules.

``` 
com.xceptance.xlt.reportgenerator.label.1.matching = .*Homepage$
com.xceptance.xlt.reportgenerator.label.1.matching.exclude = ^Goto.*
com.xceptance.xlt.reportgenerator.label.1.label = Homepage
com.xceptance.xlt.reportgenerator.label.1.stopOnMatch = false
``` 

* Matching: We match against the name only.
* Exclude: gives us the ability to specify things better than just a matching rule. 
* stopOnMatch: Allows us to stop match rule processing for that name if needed, this will simplify rule setup
* Label: The label to apply. This is just a string, so anything is permitted here.

Rules are processed in order (1, 2, ...) and you can have holes in the list (similar to mergerules) and you don't have to start with 1 or have an upper limit (well.. Integer.MAX is the limit, kind of).

jowerner:
Labeling rules are similar to merge rules, so applied the same mechanism here. Also reused the rule configuration syntax.

jowerner:
[...]

PR #507, opened 2024-07-08
#478: Labeling of Transactions, Actions, and Requests

jowerner:
* Add labeling rules (similar to merge rules)
* Try to abstract some details of rule processing (however, merge rules do not use that yet)


broeder-xceptance:
New PR for this issue: https://github.com/Xceptance/XLT/pull/651

PR #523, opened 2024-09-24
383 custom data logger

js-xc:


js-xc:
Open:
* logged data is should also be zipped in result (for download from agents and also for storing from MC)
* unit tests
* should we add a delimiter to make sure we don't blow up our machines with logged data?
* should be do any kind of csv sanity checks?
* checks on scopename and extension according to system requirements for filenames
* ...WIP

src/main/java/com/xceptance/xlt/report/providers/CustomLogsReportProvider.java
```diff
@@ -0,0 +1,201 @@
...
+        final Path targetDir = Paths.get(getConfiguration().getReportDirectory() + File.separator + CUSTOM_DATA);
+    
+        CustomLogFinder finder = new CustomLogFinder();
+        
+        finder.setBaseDir(resultsDir);
+        finder.setTargetDir(targetDir);
+        
+        try
+        {
+            Files.walkFileTree(resultsDir, finder);
```
jowerner, line 64: XLT is able to create a report also from a result archive (i.e. a zip file). That's why we usually deal with FileObjects, not Paths or Files. This means we cannot use the standard file tree walker, but have to implement that piece ourselves. See DataReaderThread or DataProcessor for inspiration how to work with FileObjects. The actual tree walker is likely just a recursive method.

src/main/java/com/xceptance/xlt/report/providers/CustomLogsReportProvider.java
```diff
@@ -0,0 +1,201 @@
...
+                // if scope/stream does not exist yet, create one and add it to list
+                if (scopeStream == null)
+                {
+                    FileOutputStream fos = new FileOutputStream(targetDir.toString() + File.separator + scopeName + ".zip");
+                    scopeStream = new ZipOutputStream(fos);
+                    foundScopes.put(scopeName, scopeStream);
+                }
+                
+                // add zip entry, copy current log file for scope    
+                scopeStream.putNextEntry(new ZipEntry(baseDir.relativize(file).toString()));
```
jowerner, line 161: Ensure the path that is passed to the ZipEntry always uses forward slashes (as required by the zip specification). Above we use code that produces a path with backslashes on Windows OS.

jowerner:
Optional feature: Before/while adding a custom log file to the zip archive, count the lines of the input file and remember the total number per scope for later display in the report as a new table column.

config/xsl/loadreport/sections/custom-values.xsl
```diff
@@ -204,6 +204,46 @@
...
+                    <xsl:for-each select="customLogs/customLog">
+	                    <xsl:sort select="scope"/>
+	                    <tr>
+	                    	<td>
+	                    		<a>
+                            		<xsl:attribute name="href"><xsl:value-of select="path"/></xsl:attribute>
+                            		<xsl:value-of select="scope"/>
+                            	</a>
+                            </td>
+	                    	<td><xsl:value-of select="format-number(size, '#,##0.###')"></xsl:value-of> B</td>
```
jowerner, line 232: Might be more user-friendly to print the size in KB, MB, etc. We have a helper template for that: `format-bytes.xsl`. See `network-table.xsl` for info on how to use that helper template. In addition, please right-align the text.

src/main/java/com/xceptance/xlt/report/providers/CustomLogsReportProvider.java
```diff
@@ -0,0 +1,217 @@
...
+     */
+    private void findLogs (FileObject file) throws IOException
+    {
+        if (file.getType() == FileType.FOLDER && !XltConstants.CONFIG_DIR_NAME.equals(file.getName().getBaseName().toString()))
+        {
+            for (final FileObject fo : file.getChildren())
+            {
+                findLogs(fo);
+            }
+        }
```
jowerner, line 138: I would like to avoid combining the file type check with another condition. Please move the second condition to a new `if` inside the body of this `if`.

src/main/java/com/xceptance/xlt/report/providers/CustomLogsReportProvider.java
```diff
@@ -0,0 +1,217 @@
...
+        return;
+    }
+
+    private void writeDataToZip(FileObject file, ZipOutputStream scopeStream) throws IOException
+    {
+        final boolean isCompressed = "gz".equalsIgnoreCase(file.getName().getExtension());
+        
+        InputStream in = isCompressed ? 
+            new GZIPInputStream(file.getContent().getInputStream(), 1024 * 16) : //TODO is my data EVER zipped? it should be, but how does it happen?
+            file.getContent().getInputStream();
```
jowerner, line 176: Custom data is not gzipped at the moment, so we can remove this code.

src/main/java/com/xceptance/xlt/report/providers/CustomLogsReportProvider.java
```diff
@@ -0,0 +1,217 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public Object createReportFragment()
+    {
+        final CustomLogsReport report = new CustomLogsReport();
+
+        FileObject results = ((ReportGeneratorConfiguration) getConfiguration()).getResultsDirectory();
+        
+        baseDir = results.getPath();
```
jowerner, line 63: When creating the report from a result archive, this won't work as the results directory is inside the archive. 

To reproduce, simply create a zip file from your results directory and pass that to the report generator.

src/main/java/com/xceptance/xlt/report/providers/CustomLogsReportProvider.java
js-xc: I thought the overall goal was to have all data zipped, so instead if removing this we should get the custom data zipped?

src/main/java/com/xceptance/xlt/api/engine/DataLogger.java
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * Adds a header to the log file. 
+     * This only makes sense for csv files and only if no data is in the file yet, 
+     * so the header may not be written if this is not the case.
+     * @param header
+     */
+    public void setHeader(String header);
+    //TODO what about format? is this freetext string or do we expect csv, i.e. a number of strings seperated 
+    //by a fixed delimiter defined seperately?
```
jowerner, line 18: I believe it is free text because we do not impose a format for the data lines either.

src/main/java/com/xceptance/xlt/api/engine/DataLogger.java
```diff
@@ -0,0 +1,34 @@
...
+    
+    /**
+     * Sets the file extension. System requirements apply.
+     * @param extension
+     */
+    public void setExtension(String extension);
+    //TODO is this csv by default or do we want other options? can user set any extension and basically save a 
+    //textfile? are there system limits as to what is allowed as a file extension (guess so)?
```
jowerner, line 26: Check with PO if he is fine with CSV as default.

src/main/java/com/xceptance/xlt/api/engine/DataLogger.java
```diff
@@ -0,0 +1,34 @@
...
+    
+    /**
+     * @param lineOfData
+     */
+    public void log(String lineOfData); 
+    //TODO is this supposed to to plausibility check, e.g. for csv?
```
jowerner, line 32: No, I believe the user is fully responsible for what is written. Maybe apart from counting lines, XLT does not evaluate the data in any way, so no checks needed.

src/main/java/com/xceptance/xlt/engine/DataManagerImpl.java
```diff
@@ -70,6 +72,8 @@ public class DataManagerImpl implements DataManager
      * Logger responsible for logging the statistics to the timer file(s).
      */
     private volatile BufferedWriter logger;
+    
+    private HashMap<String, DataLogger> dataLoggers = new HashMap<String, DataLogger>();
```
jowerner, line 76: I would feel better if this map would be a `ConcurrentHashMap`. In the rare case we have parallel activities in a single session.

Furthermore, please add some javadoc.

src/main/java/com/xceptance/xlt/engine/DataManagerImpl.java
```diff
@@ -390,4 +394,22 @@ static StringBuilder removeLineSeparators(final StringBuilder src, final char re
...
+        
+        // check if dataLogger has already been initialized
+        // otherwise create one, add it to the list and return it
+        if (dataLogger == null)
+        {            
+            dataLogger = new DataLoggerImpl(session, scope);
+            dataLoggers.put(scope, dataLogger);
+        }
+        
+        return dataLogger;
```
jowerner, line 413:
```suggestion
        return dataLoggers.computeIfAbsent(scope, sc -> new DataLoggerImpl(session, sc));
```

This is shorter and also performs some clever magic to make adding a value thread-safe because the underlying map is a `ConcurrentHashMap`.

jowerner (changes requested):

src/main/java/com/xceptance/xlt/api/engine/Session.java
```diff
@@ -115,6 +115,13 @@ public static final void logEvent(final String eventName, final String eventMess
      * @return the data manager
      */
     public abstract DataManager getDataManager();
+    
+    /**
+     * Logs a line of custom data in the datalogger of the given scope.
+     * (Shortcut method for getDataManager().dataLogger(scope).log(lineOfData))
+     *
+     */
+    public abstract void logData(String scope, String lineOfData);
```
jowerner, line 124: PO wanted `Session.logData(scope, data)` as shortcut, not `Session.getCurrent().logData(scope, data)`. Please rework this accordingly. See `Session.logEvent()` for inspiration.

In this course, please add parameter descriptions to the javadoc.

jowerner:
> Optional feature: Before/while adding a custom log file to the zip archive, count the lines of the input file and remember the total number per scope for later display in the report as a new table column.

Postponed.

Issue #527, opened 2024-10-21
Add chart to response codes that shows requests and response codes


rschwietzke:
Compose a new chart that will be part of the network page to list all response codes (ordered from 0 to ..., but only the codes we have truly seen). 

We will color the codes by range aka 0 to 99, 100 to 199, 200 to 299, and so on, so that one can easily see in what range that was.

- 1xx informational response – the request was received, continuing process
- 2xx successful – the request was successfully received, understood, and accepted
- 3xx redirection – further action needs to be taken in order to complete the request
- 4xx client error – the request contains bad syntax or cannot be fulfilled
- 5xx server error – the server failed to fulfil an apparently valid request

I imagine a chart that is similar to this, but a) colored differently, b) more compact aka not just single charts stacked.

![image](https://github.com/user-attachments/assets/34733711-685c-4648-8186-375a20dfc498)

PR #528, opened 2024-10-23
#527: Add response codes chart

jowerner:


rschwietzke (changes requested):
Sweat. Is there a way to get the numbers aligned in the front? I know that is known, but it looks odd that the response code at the vertical axis is changing is position depending on the count. How about just putting that to the right of the the chart instead?

image

jowerner:
> Sweat. Is there a way to get the numbers aligned in the front? I know that is known, but it looks odd that the response code at the vertical axis is changing is position depending on the count. How about just putting that to the right of the the chart instead?

Didn't find a way to left-align the axis labels, so we put the response code to the right side of the plots as suggested.

rschwietzke (approved):

rschwietzke:
The new chart concept with the "right axis"  text looks good.

Issue #533, opened 2024-11-14
Update Selenium


jowerner:
We need to update Selenium to get better support for newer browser versions.

PR #534, opened 2024-11-14
#533: Update Selenium

jowerner:


pom.xml
```diff
@@ -402,13 +402,13 @@
         <dependency>
             <groupId>net.bytebuddy</groupId>
             <artifactId>byte-buddy</artifactId>
-            <version>1.14.11</version>
+            <version>1.15.7</version>
```
h-arlt, line 405: Latest is v1.15.10. Upgrade?

pom.xml
```diff
@@ -71,9 +71,9 @@
         <jetty.version>9.4.53.v20231009</jetty.version>
         <log4j.version>2.22.1</log4j.version>
         <powermock.version>2.0.9</powermock.version>
-        <selenium.version>4.17.0</selenium.version>
+        <selenium.version>4.26.0</selenium.version>
         <htmlunitdriver.version>4.17.0</htmlunitdriver.version>
-        <otel.version>1.40.0</otel.version>
+        <otel.version>1.43.0</otel.version>
```
h-arlt, line 76: Latest is v1.44.1. Upgrade?

pom.xml
jowerner: 1.15.7 is the one that Selenium 2.46.0 was tested with. Didn't want to introduce additional compatibility issues.

pom.xml
jowerner: See above.

Issue #540, opened 2024-12-03
Upgrade ec2_admin to AWS Java SDK v2.x


jowerner:
`ec2_admin`, our tool for managing AWS machines, currently uses the AWS Java SDK v1.x. This version is now in maintenance mode and will be retired in December 2025. Please see https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-java-v1-x-on-december-31-2025/ for more details. Time to upgrade to the v2.x SDK.

PR #550, opened 2025-01-14
#540 Update AWS SDK to version 2.27.21

broeder-xceptance:
See https://github.com/Xceptance/XLT/issues/540

### DIFF
--- a/src/main/java/com/xceptance/xlt/ec2/AbstractEC2Client.java
+++ b/src/main/java/com/xceptance/xlt/ec2/AbstractEC2Client.java
@@ -96,14 +96,7 @@ abstract public class AbstractEC2Client
      */
     protected final AwsConfiguration awsConfiguration;
 
-    /**
-     * The "available" state constant. (Image state)
-     */
-    public static final String STATE_AVAILABLE = "available";
-
     private static final long INSTANCE_STATE_POLLING_INTERVAL = 1000;
-
-    private static final long IMAGE_STATE_POLLING_INTERVAL = 1000;
 
     /**
      * The log facility.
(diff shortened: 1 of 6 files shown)


src/main/java/com/xceptance/xlt/ec2/AbstractEC2Client.java
```diff
@@ -225,7 +237,8 @@ protected Instance waitForInstanceState(final Region region, final Instance inst
      * @throws Exception
      *             if the image didn't achieve the state within given time OR image wasn't found
      */
-    protected Image waitForImageState(final Region region, final String imageID, final String state, final long timeout) throws Exception
+    protected Image waitForImageState(final Region region, final String imageID, final ImageState state, final long timeout)
+        throws Exception
```
jowerner, line 241: Looks like this method is no longer used and can therefore be removed. In this course, we can also remove the constants STATE_AVAILABLE and IMAGE_STATE_POLLING_INTERVAL.

src/main/java/com/xceptance/xlt/ec2/AwsConfiguration.java
```diff
@@ -152,7 +151,7 @@ public AwsConfiguration(final String accessKey, final String secretKey)
             this.accessKey = StringUtils.isNotBlank(accessKey) ? accessKey : getStringProperty(PROP_ACCESS_KEY);
             this.secretKey = StringUtils.isNotBlank(secretKey) ? secretKey : getStringProperty(PROP_SECRET_KEY);
 
-            protocol = Protocol.valueOf(getStringProperty(PROP_PROTOCOL, "https").toUpperCase());
+            protocol = getStringProperty(PROP_PROTOCOL, "https");
```
jowerner, line 154: Since the protocol is now relevant only for the proxy, I would like to change things a bit:

* Rename the constant `PROP_PROTOCOL` to `PROP_PROXY_PROTOCOL`. Adjust its value accordingly.
* Rename the property `aws.protocol` to `aws.proxy.protocol` in the `ec2_admin.properties` file. Also move it into the proxy section.
* Change the default value of the proxy protocol from https to http.

jowerner (changes requested):
See comments.

src/main/java/com/xceptance/xlt/ec2/AbstractEC2Client.java
broeder-xceptance: Done.

src/main/java/com/xceptance/xlt/ec2/AwsConfiguration.java
broeder-xceptance: Done.

jowerner (approved):

PR #556, opened 2025-02-12
fix

js-xc:


resultbrowser/src/js/resultbrowser.js
```diff
@@ -547,7 +547,7 @@
 
     function isJsonContent(contentType) {
         //  e.g. "application/json" or "application/<...>+json"
-        return /^application\/(.+\+)?json$/.test(contentType);
+        return /^application\/(.+\+)?json/.test(contentType);
```
h-arlt, line 550: This change might fix the issue but introduces another one as we will treat content as JSON that might not be JSON, e.g. [application/jsonpath](https://www.iana.org/assignments/media-types/application/jsonpath).

Or in other words: the regex was too strict before and is too lax now :smile: 

What about to keep the regex but sanitize the input instead?
```suggestion
        contentType = (contentType || "").split(";")[0].trim();
        return /^application\/(.+\+)?json$/.test(contentType);
```

resultbrowser/src/js/resultbrowser.js
```diff
@@ -546,8 +546,8 @@
     }
 
     function isJsonContent(contentType) {
-        //  e.g. "application/json" or "application/<...>+json"
-        return /^application\/(.+\+)?json/.test(contentType);
+        //  e.g. "application/json" or "application/<...>+json" or "application/json;charset=utf-8"
+        return /^application\/(.+\+)?json(;.*)?$/.test(contentType);
```
h-arlt: Maybe I'm paranoid but better safe than sorry: to ensure that neither leading nor ending whitespace characters break this function, we may trim the input string using our `trim` function declared above. Additionally, the separator character ';' may be preceeded by some whitespace characters as well.

```suggestion
        return /^application\/(.+\+)?json(\s*;.*)?$/.test(trim(contentType));
```

PR #561, opened 2025-02-19
load profile percentages (WIP)

js-xc:


config/xsl/common/util/load-profile-table.xsl
```diff
@@ -262,6 +264,11 @@
                                     </xsl:if>
                                     <xsl:value-of select="numberOfUsers"/>
                                 </td>
+                                <td class="value number">
+                                    <xsl:text>(</xsl:text>
+                                    <xsl:value-of select="format-number(numberOfUsersPercentage, '#,##0.00')" />
+                                    <xsl:text>%)</xsl:text>
```
jowerner, line 270: Don't enclose the percentage value in parentheses if we display it in a separate column.

src/main/java/com/xceptance/xlt/report/providers/ConfigurationReportProvider.java
```diff
@@ -138,6 +142,46 @@ public Object createReportFragment()
...
+            int[][] arrivalRateFunction = loadProfileConfig.getArrivalRate();
+            loadProfileConfig.setArrivalRatePercentage(ReportUtils.calculatePercentage(getMaxFromLoadFunction(arrivalRateFunction), arrivalRateTotal));
+            int[][] userCountFunction = loadProfileConfig.getNumberOfUsers();
+            loadProfileConfig.setNumberOfUsersPercentage(ReportUtils.calculatePercentage(getMaxFromLoadFunction(userCountFunction), userCountTotal));
+            
+        }        
+        return;
+    }
+    
+    private int getMaxFromLoadFunction(int[][] loadFunction)
```
jowerner, line 170: If an arrival rate is not defined for a scenario, the load function is null, so this method will produce a NPE.

config/xsl/common/util/load-profile-table.xsl
```diff
@@ -292,6 +299,11 @@
                                         </xsl:otherwise>
                                     </xsl:choose>
                                 </td>
+                                <td class="value number">
+                                	<xsl:text>(</xsl:text>
+                                    <xsl:value-of select="format-number(arrivalRatePercentage, '#,##0.00')" />
```
jowerner, line 304: PO wanted to display an `&ndash;` if no arrival rate was defined.

jowerner (changes requested):

js-xc:
open: copy data from `TestCaseLoadProfileConfiguration` to display class, only add percentages there

jowerner:
When creating a report from an empty results folder, the load profile table now looks broken as the body and footer row miss two columns each.

jowerner (approved):

Issue #546, opened 2025-01-04
Extend API return config and data directory as Path


rschwietzke:
In `config` and `config/data`, we typically store test data as well as other files. There is right now no way for an XLT user to programmatically get these locations without accessing properties and making the path up themselves.

Extend `XltProperties` to return a `Path` to config as well as a `Path` to `config/data`.

* XltProperties.getConfigDirectory()
* XltProperties.getDataDirectory()

PR #562, opened 2025-02-26
#546: Extend XltProperties API to return config and data directory

broeder-xceptance:
See: https://github.com/Xceptance/XLT/issues/546


src/main/java/com/xceptance/xlt/util/XltPropertiesImpl.java
```diff
@@ -216,6 +227,9 @@ public XltPropertiesImpl(final Properties properties)
      */
     public XltPropertiesImpl(final Optional<Properties> properties)
     {
+        this.configDirectory = XltExecutionContext.getCurrent().getTestSuiteConfigDir().getPath();
+        this.dataDirectory = this.configDirectory.resolve(XltConstants.DATA_DIR_NAME);
```
jowerner, line 231: Note that the data directory can also be given via property "com.xceptance.xlt.data.directory". Only if this is not set, should we fall back to "config/data".

src/main/java/com/xceptance/xlt/util/XltPropertiesImpl.java
```diff
@@ -241,6 +255,9 @@ private synchronized void initialize(final FileObject homeDirectory, final FileO
         var cd = configDirectory == null ? XltExecutionContext.getCurrent().getTestSuiteConfigDir() : configDirectory;
         loadProperties(hd, cd, ignoreMissingIncludes, staySilent);
 
+        this.configDirectory = cd.getPath();
+        this.dataDirectory = this.configDirectory.resolve(XltConstants.DATA_DIR_NAME);
```
jowerner, line 259: See above.

src/test/java/com/xceptance/xlt/api/util/XltPropertiesTest.java
```diff
@@ -696,4 +703,49 @@ public void includeConfigProblem() throws FileSystemException
...
+    public void getConfigAndDataDirs_EmptyProperties() throws FileSystemException
+    {
+        setup("propertytest_hp", "propertytest_hp/config");
+
+        var p = new XltPropertiesImpl();
+
+        assertEquals(configDir.getPath(), p.getConfigDirectory());
+        assertEquals(configDir.getPath().resolve(XltConstants.DATA_DIR_NAME), p.getDataDirectory());
+    }
 }
```
jowerner, line 751: We need more tests in case the data directory is redefined via property.

jowerner (changes requested):
Now that we have those new methods, we might also use them in our code, for example in DataProvider.

src/main/java/com/xceptance/xlt/util/XltPropertiesImpl.java
broeder-xceptance: Done.

src/main/java/com/xceptance/xlt/util/XltPropertiesImpl.java
broeder-xceptance: Done.

src/test/java/com/xceptance/xlt/api/util/XltPropertiesTest.java
broeder-xceptance: Done.

jowerner (approved):

Issue #580, opened 2025-04-17
Parse error for URLs with an IPv6 address


jowerner:
When removing the user info from URLs, XLT fails to parse URLs where the host part is given as an IPv6 address, such as in `https://[2600:1900:4000:e878:0:1:0:0]:8443/`:

```
java.lang.NumberFormatException: For input string: "1900:4000:e878:0:1:0:0]:8443"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
	at java.base/java.lang.Integer.parseInt(Integer.java:662)
	at java.base/java.lang.Integer.parseInt(Integer.java:778)
	at com.xceptance.xlt.engine.util.UrlUtils.parseUrlString(UrlUtils.java:158)
	at com.xceptance.xlt.engine.util.UrlUtils.removeUserInfo(UrlUtils.java:239)
	at com.xceptance.xlt.engine.util.URLCleaner.removeUserInfoIfNecessaryAsString(URLCleaner.java:69)
	at com.xceptance.xlt.engine.XltHttpWebConnection.cleanHeaders(XltHttpWebConnection.java:510)
```

Workaround: Set property `com.xceptance.xlt.results.data.request.removeUserInfoFromURL = false`.

PR #595, opened 2025-07-01
#580 fix parse error for IPv6 addresses

broeder-xceptance:
See: https://github.com/Xceptance/XLT/issues/580

### DIFF
--- a/src/main/java/com/xceptance/xlt/engine/util/UrlUtils.java
+++ b/src/main/java/com/xceptance/xlt/engine/util/UrlUtils.java
@@ -18,13 +18,14 @@ package com.xceptance.xlt.engine.util;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.client.utils.URLEncodedUtils;
 import org.htmlunit.httpclient.HttpClientConverter;
 import org.htmlunit.util.NameValuePair;
 
-import com.xceptance.common.util.RegExUtils;
 import com.xceptance.xlt.common.XltConstants;
 
 /**
@@ -106,8 +107,10 @@ public final class UrlUtils
      * @param urlString
      *            the input URL
      * @return a {@link URLInfo} object with the details
+     * @throws NumberFormatException
+     *             if the port of the URL can't be parsed as an Integer
      */
-    public static URLInfo parseUrlString(final String urlString)
+    public static URLInfo parseUrlString(final String urlString) throws NumberFormatException
     {
         if (StringUtils.isEmpty(urlString))
         {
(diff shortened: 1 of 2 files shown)


src/main/java/com/xceptance/xlt/engine/util/UrlUtils.java
```diff
@@ -148,15 +151,24 @@ public static URLInfo parseUrlString(final String urlString)
...
         hostPort = parts[1];
 
-        parts = split(hostPort, ":", true);
-        host = parts[0];
-        port = parts[1];
-
-        int p = -1;
-        if (StringUtils.isNotBlank(port) && RegExUtils.isMatching(port, "\\d+"))
+        // check if the host is in IPv6 format
+        final Matcher ipv6Matcher = Pattern.compile("^(\\[.*\\])(?::(\\d+))?$").matcher(hostPort);
```
jowerner, line 155: Don't compile the regex into a pattern each time, but do this only once and store the result as a constant in a static field.

Furthermore, for an input of `[::1]:foo`, the code would decide that this must be an IPv4 address, but it is rather a valid IPv6 address with an invalid port. How about changing the regex a little so that the port part can be whatever. The code to parse the port would then throw a NFE with a more meaningful message.

src/test/java/com/xceptance/xlt/engine/util/UrlUtilsTest.java
```diff
@@ -66,7 +66,37 @@ public void testParseUrl_ProtoHost() throws Throwable
...
+        Assert.assertNotNull("Failed to parse url", info);
+
+        Assert.assertEquals("http", info.getProtocol());
+        Assert.assertNull(info.getUserInfo());
+        Assert.assertEquals("[2001:db8:1111:2222:aaaa:BBBB:1a2b:cd34]", info.getHost());
+        Assert.assertEquals(-1, info.getPort());
+        Assert.assertEquals("", info.getPath());
+        Assert.assertNull(info.getQuery());
+        Assert.assertNull(info.getFragment());
+    }
```
jowerner, line 96: We should avoid duplicating the tests for host name, IPv4 address, and IPv6 address inputs. Instead, we could run the same test with different inputs. In JUnit4, there is the concept of parameterized tests. There is also an alternative approach. See OkHttpRequestBodyEncodingTest for an example.

jowerner (changes requested):

src/main/java/com/xceptance/xlt/engine/util/UrlUtils.java
broeder-xceptance: Done.

src/test/java/com/xceptance/xlt/engine/util/UrlUtilsTest.java
broeder-xceptance: Done.

PR #605, opened 2025-09-05
#604: Implausible results when filtering a request table

### LINKED ISSUE #604: Implausible results when filtering a request table
When filtering the entries of the request table in a load test report, the filter result is not always plausible. For example, when searching for "ddd", the request "SomeActionName (aaaaaaaaaa) (bbbbbbbbbb) (cccccccccc) (dddddddddd)" might still be filtered out, even though it clearly contains "ddd".

jowerner:
When cutting off the URL part before filtering, make sure we don't eliminate parts of the request name by accident.

### DIFF
--- a/config/testreport/js/xlt.js
+++ b/config/testreport/js/xlt.js
@@ -40,7 +40,7 @@
     // the filter function, returns true if the value is to be shown
     function doFilter(value, filterPhrase) {
         // request table cells contain the URLs as well, so cut them off
-        value = value.trim().split('\n')[0];
+        value = value.trim().replace(/\s*[0-9,.]+ distinct URL.+$/s, "");
 
         // split filter phrase into filters
         var filters = filterPhrase.split('|');
--- a/config/testreport/js/xlt.js
+++ b/config/testreport/js/xlt.js
@@ -39,8 +39,7 @@
 
     // the filter function, returns true if the value is to be shown
     function doFilter(value, filterPhrase) {
-        // request table cells contain the URLs as well, so cut them off
-        value = value.trim().replace(/\s*[0-9,.]+ distinct URL.+$/s, "");
+        value = value.trim();
 
         // split filter phrase into filters
         var filters = filterPhrase.split('|');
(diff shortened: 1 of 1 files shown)


config/testreport/js/xlt.js
```diff
@@ -40,7 +40,7 @@
     // the filter function, returns true if the value is to be shown
     function doFilter(value, filterPhrase) {
         // request table cells contain the URLs as well, so cut them off
-        value = value.trim().split('\n')[0];
+        value = value.trim().replace(/\s*[0-9,.]+ distinct URL.+$/s, "");
```
h-arlt: What about to make the regex more flexible with respect to the number of space characters between the several parts?
```suggestion
        value = value.trim().replace(/\s*[0-9,.]+\s+distinct\s+URL.+$/s, "");
```

If not for all, I'd suggest to do so for the space characters that separate the (localized) number and the word 'distinct' at least.

Anyway, the used approach only works as long as request name does not contain the given pattern; otherwise, we'd cut it off as well. What I'm trying to say is: the new approach fixes the issue for the specific use-case but there might be still use-cases where we cut off parts from the request name.

h-arlt: Instead of fiddling around with the computed text in our filter function (where we cannot say for sure which parts of the text we can safely cut off), we may patch the function `getCellValue` of the Table plugin such that it "skips" elements having CSS class `cluetip-data`. This way, we can safely omit any text inside tooltip containers for sorting and filtering. 

This patch can be installed by function `setupTables`:
```js
    (function setupTables() {
         const tableGetCellValue = Table.getCellValue;
         Table.getCellValue = function() {
             const node = arguments[0];
             // don't let element nodes having CSS class 'cluetip-data' contribute any text to computed table cell value
             if (node && node.nodeType === 1 && node.classList.contains("cluetip-data")) {
                 return "";
             }
             // invoke the original function
             return tableGetCellValue.apply(Table, arguments);
         };    
         Table.auto();
         ...
```

Once the patch is in place, our filter function just need to trim the passed value and that's it.

jowerner: Done.

h-arlt (approved):

PR #607, opened 2025-09-10
#606: Custom cookies may not be working with OkHttp

### LINKED ISSUE #606: Custom cookies may not be working with OkHttp
Sometimes, test suites need to inject custom cookies. This is done by creating a Cookie instance and adding it to HtmlUnit's cookie manager.

However, the HTTP clients we support (currently Apache or OkHttp) have different requirements for the cookie details that must be provided. For example, Apache tolerates a missing cookie path, but OkHttp requires it and otherwise throws an NPE:

```
java.lang.NullPointerException: Parameter specified as non-null is null: method okhttp3.Cookie$Builder.path, parameter path
	at okhttp3.Cookie$Builder.path(Cookie.kt) ~[okhttp-4.12.0.jar:?]
	at com.xceptance.xlt.engine.htmlunit.okhttp3.CookieJarImpl.toOkHttpCookie(CookieJarImpl.java:102) ~[xlt-9.0.0.jar:9.0.0]
```

Since this only happens later, when the cookies are actually sent, this exception is not helpful for determining the root cause. Perhaps we can provide more context information here.

jowerner:
* Sanitize values of the passed HtmlUnit cookie for use with OkHttp:
    * domain: must not be empty
    * name: must be non-null and trimmed
    * value: must be non-null and trimmed
    * path: must be non-null and start with "/"
    * expiresAt: don't set if null
* Add unit tests to check correct behavior
* Extended JUnitParamsUtils and added unit tests for it

### DIFF
--- a/src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/CookieJarImpl.java
+++ b/src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/CookieJarImpl.java
@@ -19,7 +19,6 @@ import java.util.Date;
 import java.util.List;
 import java.util.stream.Collectors;
 
-import org.apache.commons.lang3.StringUtils;
 import org.htmlunit.CookieManager;
 import org.htmlunit.util.Cookie;
 
@@ -32,9 +31,7 @@ import okhttp3.HttpUrl;
  */
 class CookieJarImpl implements CookieJar
 {
-    static final String ERROR_MSG_DOMAIN = "Domain not set in cookie";
-
-    static final String ERROR_MSG_PATH = "Path not set or does not start with '/' in cookie";
+    static final String ERROR_MSG_COOKIE_INVALID = "Cookie not valid for use with OkHttp";
 
     /**
      * HtmlUnit's cookie manager.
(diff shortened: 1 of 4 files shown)


src/test/java/com/xceptance/xlt/engine/htmlunit/okhttp3/CookieJarImplTest.java
```diff
@@ -0,0 +1,155 @@
...
+    {
+        try
+        {
+            CookieJarImpl.toOkHttpCookie(huCookie);
+            Assert.fail("Expected IllegalArgumentException");
+        }
+        catch (final IllegalArgumentException e)
+        {
+            Assert.assertTrue("Unexpected exception message", e.getMessage().startsWith(expectedMessagePrefix));
+        }
```
broeder-xceptance: This might be a cleaner solution for this:

```suggestion
        final IllegalArgumentException exception = Assert.assertThrows(IllegalArgumentException.class, () -> CookieJarImpl.toOkHttpCookie(huCookie));
        MatcherAssert.assertThat(exception.getMessage(), CoreMatchers.startsWith(expectedMessagePrefix));
```

(The MatcherAssert has a more helpful error message in case of failure, since it shows the actual and expected values)

src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/CookieJarImpl.java
```diff
@@ -95,24 +99,50 @@ private static Cookie toHtmlUnitCookie(final okhttp3.Cookie okHttpCookie)
...
-        builder.name(htmlUnitCookie.getName()).value(htmlUnitCookie.getValue()).path(htmlUnitCookie.getPath())
-               .expiresAt(htmlUnitCookie.getExpires().getTime());
+        final String domain = StringUtils.defaultString(htmlUnitCookie.getDomain()).trim();
+        if (!domain.isEmpty())
+        {
+            if (domain.startsWith("."))
+            {
+                // "wildcard" domain, such as ".example.com"
+                // the cookie builder expects the domain to be passed without the leading dot
+                builder.domain(StringUtils.stripStart(domain, "."));
```
broeder-xceptance: In the edge case that the domain value is exactly `"."` this would result in `builder.domain("")` and throw an exception. Should we handle that case as well?

src/test/java/com/xceptance/xlt/engine/htmlunit/okhttp3/CookieJarImplTest.java
```diff
@@ -0,0 +1,155 @@
...
+        Assert.assertEquals("", okCookie.name());
+    }
+
+    @Test
+    @Parameters(source = NameProvider.class)
+    public void toOkHttpCookie_ValidName(final String name)
+    {
+        final var okCookie = CookieJarImpl.toOkHttpCookie(new Cookie("domain", name, "value", "/", null, false, false));
+
+        Assert.assertEquals("name", okCookie.name());
```
broeder-xceptance: This works, but maybe we should change the validation to this:

```suggestion
        Assert.assertEquals(name.trim(), okCookie.name());
```

So the test would still succeed if we ever change or expand the test data set.
(The same goes for the ValidValue and ValidPath tests)

src/test/java/com/xceptance/xlt/engine/htmlunit/okhttp3/CookieJarImplTest.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    @Test
+    @Parameters(source = BlankStringOrNullParamProvider.class)
+    public void toOkHttpCookie_BlankValue(final String value)
```
broeder-xceptance: For consistency, maybe rename this test to `toOkHttpCookie_BlankValueOrNull` (same for `toOkHttpCookie_BlankPath`).

src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/CookieJarImpl.java
```diff
@@ -95,24 +99,50 @@ private static Cookie toHtmlUnitCookie(final okhttp3.Cookie okHttpCookie)
...
         {
-            // "host" domain, such as "www.example.com"
-            builder.hostOnlyDomain(domain);
+            throw new IllegalArgumentException(ERROR_MSG_PATH + ": '" + htmlUnitCookie + "'");
+        }
+
+        final Date expires = htmlUnitCookie.getExpires();
+        if (expires != null)
+        {
+            builder.expiresAt(expires.getTime());
```
broeder-xceptance: There probably isn't much that could go wrong here, but for the sake of completeness should we maybe add unit tests for null and non-null expiry dates?

broeder-xceptance (changes requested):

broeder-xceptance (approved):

broeder-xceptance (approved):

PR #610, opened 2025-09-23
608 integration

### LINKED ISSUE #608: Add the ability to specify url merge rules by text only
Because modern applications have a lot of urls with random data, the merge rule cache is often not effective. It might be beneficial to also have the ability to check by text instead of by regexp or combine text and regexp. In the latter case, we do a cheap indexOf check first and go the expensive regex route only if needed.

rschwietzke:
Addresses #608, some incorrect branch history in the backpack....


jowerner (approved):

jowerner:
@rschwietzke I had to make some changes to RequestData to make the unit tests green again. However, these changes may harm performance so please have a look at https://github.com/Xceptance/XLT/pull/610/commits/da7bc3bb899b829729c2110c81c61932109b12f2.

rschwietzke:
The change is good and does not harm performance, seems to even improve it.

PR #612, opened 2025-10-06
feat: Add jitter to load test setup

### LINKED ISSUE #545: Apply Jitter to Load Test Setup
## Non-Goal
We would rather not create something fully random or partially intelligent. Keep in mind, that the setup is precalculated.

rschwietzke:
This commit introduces a new 'jitter' feature to the load test configuration, allowing for more realistic load simulations.

A new 'jitter' property can be added to the test case configuration, which applies random variations to the number of users or arrival rate over time.

The jitter is calculated using a Gaussian distribution and is seeded for reproducibility.

The implementation ensures that supporting points in a flexible load profile remain unchanged.

Unit tests have been added to validate the new functionality.

Fixes #545

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -31,7 +31,7 @@ resultbrowser.dir.source = resultbrowser/dist
 resultbrowser.dir.target = ${classes.dir}/com/xceptance/xlt/engine/resultbrowser/assets
 
 # Linux
-timerrecorder.chrome.executable = chromium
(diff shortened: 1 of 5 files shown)


rschwietzke:
That is once again a Jules test. So we have to carefully review the suggestion and learn how much is useful or not. Don't merge that thing right away.

rschwietzke:
We are going to try that again. Closing as outdated.

PR #619, opened 2025-10-23
#615: XLT keeps sending the Authorization header

### LINKED ISSUE #615: XLT keeps sending the Authorization header
After setting a custom Authorization header programmatically at the web client and making the first request, XLT continues to send that header for all subsequent requests in the same session, even if the header has been removed again from the web client. This only happens when using OkHttp as the underlying HTTP client.

jowerner:
AuthenticatorImpl
* moved authentication cache functionality here
* implement reactive (proxy) authorization by consulting the credentials provider, but also cache the result
* support preemptive (proxy) authorization by providing access to the cached (proxy) authorization headers, but ensure that the previously used credentials are still present in the credentials provider

AuthorizationHeaderInterceptor
* do no longer put authorization headers already present at the request into the cache
* implement preemptive authorization by asking the authenticator for cached authorization headers
* handle authorization for proxies as well

OkHttp3WebConnection
* make the authenticator global as it now contains the cache


src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/AuthenticatorImpl.java
```diff
@@ -35,11 +38,51 @@
...
  * An OkHttp {@link Authenticator} that is backed by HtmlUnit's credentials provider.
+ *
+ * @see AuthorizationHeaderInterceptor
  */
 class AuthenticatorImpl implements Authenticator
 {
+    /**
+     * The result of an authentication attempt ready to be stored in the authentication cache.
+     */
+    static class AuthResult
```
h-arlt: *LOW* Looks like this is a perfect candidate for being implemented as record. WDYT?

src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/AuthenticatorImpl.java
```diff
@@ -123,11 +165,137 @@ else if (isProxyAuth)
...
     {
-        final String basicAuthHeaderName = isProxyAuth ? HttpRequestHeaders.PROXY_AUTHORIZATION : HttpRequestHeaders.AUTHORIZATION;
-        final String basicAuthHeaderValue = okhttp3.Credentials.basic(credentials.getUserPrincipal().getName(), credentials.getPassword());
+        final AuthScope authScope = createAuthScope(challenge, route, isProxyAuth);
 
-        return response.request().newBuilder().header(basicAuthHeaderName, basicAuthHeaderValue).build();
+        final Credentials credentials = credentialsProvider.getCredentials(authScope);
+        if (credentials != null)
+        {
+            final String basicAuthHeaderName = isProxyAuth ? HttpRequestHeaders.PROXY_AUTHORIZATION : HttpRequestHeaders.AUTHORIZATION;
```
h-arlt: This may be extracted into a method - e.g. `getAuthHeaderName(boolean isProxyAuth)`  - that would come in handy at lines 101-103 as well. This method may be even provided by a utilty class (if such one exists yet).

src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/AuthenticatorImpl.java
```diff
@@ -65,27 +108,15 @@ public AuthenticatorImpl(final CredentialsProvider credentialsProvider)
         // check each provided challenge
         for (final Challenge challenge : response.challenges())
         {
-            // a special scheme that OkHttp uses for preemptive HTTPS proxy authentication
+            // preemptive basic authentication for HTTPS proxies using a special scheme
```
h-arlt: The previous comment gave more insight IMO.

What about to change it as follows:
```suggestion
            // preemptive basic authentication for HTTPS proxies using a special scheme used by OkHttp only
```

src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/OkHttp3WebConnection.java
```diff
@@ -84,7 +83,7 @@ public class OkHttp3WebConnection extends AbstractWebConnection<OkHttpClient, Re
 
     private static final SSLSocketFactory INSECURE_SSL_SOCKET_FACTORY = createInsecureSslSocketFactory();
 
-    private final AuthenticationCache authenticationCache;
+    private AuthenticatorImpl authenticator;
```
h-arlt: Field is written to in c'tor only. Thus, I'd suggest to delcare it as `final`.

src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/AuthenticatorImpl.java
jowerner: Done.

src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/AuthenticatorImpl.java
jowerner: Done.

src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/AuthenticatorImpl.java
jowerner: Done.

src/main/java/com/xceptance/xlt/engine/htmlunit/okhttp3/OkHttp3WebConnection.java
jowerner: Done.

PR #620, opened 2025-10-23
#614: Provide dynamic zoomable charts with labels/displaying chart values

### LINKED ISSUE #614: Provide dynamic zoomable charts with labels/displaying chart values
## Goal
We want to enhance the reporting intelligence by adding dynamic charts that can be used to zoom in on the results as well as provide data per label aka datapoint, when hovering with the mouse.

## Non-Goal
The zoomable component will not change the resolution, just enable the user to see more of the otherwise very compressed data points. Also, only the main charts will be alternatively available in this enhanced version. 

It is not needed to be able to copy the image easily, aka as PNG. If you need that chart, you have to screenshot it.

## Requirements
We are looking for something like that: 

Image
https://echarts.apache.org/examples/en/editor.html?c=intraday-breaks-1

Because we capture min/max/avg and a rough distribution for the second we captured, for this chart we have these options and might have to try them out:

1. Use only dots and paint all the datapoints, the average is a linie
1. Use only min, max and avg and draw three lines, the max and min are dimmed, so one sees them but they are not dominating
1. Do the previous but use a band min-max as the background so one sees the overall width but is not distracted

When hovering over, for each x-resolution step, one can see min, max, avg and potentially the x-axis value as headline or something like that.

### No Display By Default
When the feature is enabled, one still has to either click on the chart to open a popover that shows the dynamic one, or we put the dynamic one on another tab. But it should only become active when it is viewed to avoid loading time. Popover dialogs might be preferred due to the larger areas it can occupy.

The feature can be disabled in the report generator properties, but is on by default.

### No Extra Resolution 
We are not increasing the resolution at the moment, that might come later aka the data the original chart provides is used.

jowerner:
* store basic runtime chart data to an additional JSON file
* add a new "Dynamic Overview" tab to the chart section containing a zoomable Apache echart showing the data from the JSON file
* add a property to enable/disable dynamic charts


config/testreport/js/xlt.js
```diff
@@ -564,6 +564,231 @@
...
+                            xhr.overrideMimeType("application/json");
+                        },
+                        success: function (data) {
+                            echart.hideLoading();
+
+                            // create time series data
+                            var dataMean = data.map(item => [item[0], item[1]]);                    // timestamp and mean value
+                            var dataMinimum = data.map(item => [item[0], item[2]]);                 // timestamp and min value
+                            var dataMaximum = data.map(item => [item[0], item[3]]);                 // timestamp and max value
+                            var dataMaxMinDiff = data.map(item => [item[0], item[3] - item[2]]);    // timestamp and diff value
```
h-arlt: Instead of iterating over the (potentially large amount of) data multiple times, we may optimize this a little bit and iterate over it only once using a simple for-of loop.

```suggestion
                            var dataMean = [];
                            var dataMinimum = [];
                            var dataMaximum = [];
                            var dataMinMaxDiff = [];

                            // create time series data
                            for (var item of data) {
                                // timestamp and mean value
                                dataMean.push([item[0], item[1]]);
                                // timestamp and min value
                                dataMinimum.push([item[0], item[2]]);
                                // timestamp and max value
                                dataMaximum.push([item[0], item[3]]);
                                // timestamp and diff value
                                dataMaxMinDiff.push([item[0], item[3] - item[2]]);
                            }
```

src/main/java/com/xceptance/xlt/report/ReportGeneratorConfiguration.java
```diff
@@ -223,6 +223,10 @@ public enum ChartCappingMode
     
     private static final String PROP_CUSTOM_DATA_AGGREGATE_FILE_CONTENTS = PROP_PREFIX + "customDataLogs.aggregateFileContents";
 
+    private static final String PROP_DYNAMIC_CHARTS_PREFIX = PROP_PREFIX + "dynamicCharts.";
```
h-arlt: *LOW* IMO this constant is not really necessary as it is used only once.

config/testreport/js/xlt.js
jowerner: Done.

src/main/java/com/xceptance/xlt/report/ReportGeneratorConfiguration.java
jowerner: Done.

PR #621, opened 2025-10-24
#616: Result Browser: Show body for all response types

### LINKED ISSUE #616: Result Browser: Show body for all response types
Currently, the response body of certain types of responses are hidden. For example, this is done for redirect responses. The bodies of these responses are usually empty. However, it is legitimate to send data in the response body, so the Result Browser should display it.

jowerner:
If there is a response file for a request, show its content in the Response Body tab, no matter what the status code or the content-length header was. Seems that loading empty files no longer cause issues.

### DIFF
--- a/resultbrowser/src/js/resultbrowser.js
+++ b/resultbrowser/src/js/resultbrowser.js
@@ -218,7 +218,7 @@
                 }
                 setText(requestText, s);
             });
-            
+
         	const actionimgSwitch = getElementById("actionimgSwitch");
         	actionimgSwitch.addEventListener("click", function () {
                 zoomImage(actionimgSwitch);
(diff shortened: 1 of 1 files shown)


resultbrowser/src/js/resultbrowser.js
```diff
@@ -749,23 +749,7 @@
...
-            if (/30[0-8]|20[45]/.test(respCode) || !rqData.fileName) {
-                return true;
-            }
-
-            for (let i = 0, l = headers.length, h; i < l; i++) {
-                h = headers[i];
-                if (h.name_ === "Content-Length") {
-                    return h.value_ === "0";
-                }
-            }
```
h-arlt, line 766: Isn't this still legit? In case XLT omits empty responses, we don't need this check. Otherwise, we may keep it to avoid attempts to load an empty file and show, well, no content.

PR #623, opened 2025-10-24
#309: Enable the average report to display more than one average

### LINKED ISSUE #309: Enable the average report to display more than one average

broeder-xceptance:
See: https://github.com/Xceptance/XLT/issues/309


src/main/java/com/xceptance/xlt/report/util/JFreeChartUtils.java
```diff
@@ -395,6 +428,18 @@ public static JFreeChart createAverageLineChart(final String seriesName, final S
             meanSeries.add(lastItem.getPeriod(), mean);
         }
 
+        // add additional averages if there are any
+        if (additionalAverageValueSeriesList != null)
+        {
+            // don't add more than the allowed maximum of additional averages
+            final int maxAdditionalAveragesCount = Math.min(additionalAverageValueSeriesList.size(),
```
h-arlt, line 435: Boundary checks and other validations are already performed when parsing report configuration. IMO JFreeUtils should just process what was given.

broeder-xceptance: For the current scenario, the validations aren't necessarily needed, yes.

However, there's the argument to be made that a utils class shouldn't assume the context it is used in. If we have another use case in a few months and call these same methods with input parameters that weren't pre-validated, we might run into issues.

I'd say the validation in the report generator configuration is more about a value range we define in that use case and about being user-friendly, since we can actually point the user to the exact properties they have to fix at this point. The validation in the JFreeChartUtils is more about the technical limitations of the utils class.

This specific validation (the number of averages per chart) als stems from the decision to limit the max number of lines (and available colors), because things get too cluttered after a certain point. So "no more than 5 lines" is a limitation of the graph and should be independent of the input data.

Does that make sense?

h-arlt: Absolutely. Thanks for clarification :+1:

src/test/java/com/xceptance/common/util/ParseUtilsTest.java
```diff
@@ -466,6 +467,35 @@ public void parseLongFallback_NoLong() throws ParseException
         Assert.assertTrue(ParseUtils.parseLong("10,aa", 10) == 10);
     }
 
+    @Test
+    public void parseIntPercentage() throws ParseException
+    {
+        Assert.assertEquals(25, ParseUtils.parseIntPercentage("25%"));
+        Assert.assertEquals(25, ParseUtils.parseIntPercentage("25"));
+
```
jowerner, line 476:
```suggestion

        Assert.assertEquals(25, ParseUtils.parseIntPercentage("25 %"));

```

config/reportgenerator.properties
```diff
@@ -118,8 +118,49 @@ com.xceptance.xlt.reportgenerator.charts.height = 300
...
+###############################################################################
+
+com.xceptance.xlt.reportgenerator.charts.commonAverage.type = percentage
+com.xceptance.xlt.reportgenerator.charts.commonAverage.value = 5%
+
+com.xceptance.xlt.reportgenerator.charts.averages.1.type = percentage
+com.xceptance.xlt.reportgenerator.charts.averages.1.value = 2%
+
+com.xceptance.xlt.reportgenerator.charts.averages.2.type = percentage
+com.xceptance.xlt.reportgenerator.charts.averages.2.value = 10%
```
jowerner, line 211: It is good to have a new section with a useful help text. However, now the following properties seem to belong to this section, which is not true. We need to regroup the properties and maybe introduce new sections.

src/test/java/com/xceptance/common/util/ParseUtilsTest.java
broeder-xceptance: Done.

config/reportgenerator.properties
broeder-xceptance: Done.

PR #630, opened 2025-11-19
#628: Slower page load times due to dynamic charts

### LINKED ISSUE #628: Slower page load times due to dynamic charts
Since dynamic charts were introduced in the load test report, the loading times for report pages have increased. For instance, if the Requests page contains hundreds of requests, it may become unresponsive for 10 seconds or more.

Setting up the dynamic charts takes time, but we can postpone that task until the Dynamic Overview tab is opened.

jowerner:
Initialize the echart only when opening the Dynamic Overview tab for the first time.

### DIFF
--- a/config/testreport/js/xlt.js
+++ b/config/testreport/js/xlt.js
@@ -567,7 +567,7 @@
         // lazy load the dynamic echarts to speed up the site
         (function setupECharts() {
             // Creates the echart, loads the chart data, and completes the echart setup.
-            function setUpEChart(echartDiv) {
+            function initAndLoadEChart(echartDiv) {
                 var name = echartDiv.getAttribute('name');
                 var url = echartDiv.getAttribute('src');
 
@@ -686,7 +686,7 @@
                                     }
                                 },
                                 axisLabel: {
-                                        fontSize: 10
+                                    fontSize: 10
                                 },
                                 splitLine: {
                                     show: true,
@@ -710,7 +710,7 @@
                                     }
                                 },
                                 axisLabel: {
-                                        fontSize: 10
+                                    fontSize: 10
                                 },
                                 splitLine: {
                                     show: true,
@@ -809,7 +809,7 @@
 
                     // add a one-time click handler to the current tab that sets up the echart
                     $(this).one("click", function() {
-                        setUpEChart(echartDiv);
+                        initAndLoadEChart(echartDiv);
                     });
                 });
             });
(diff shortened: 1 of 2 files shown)


config/testreport/js/xlt.js
```diff
@@ -564,239 +564,254 @@
...
-                                dataMean.push([item[0], item[1]]);
-                                // timestamp and min value
-                                dataMinimum.push([item[0], item[2]]);
-                                // timestamp and max value
-                                dataMaximum.push([item[0], item[3]]);
-                                // timestamp and diff value
-                                dataMaxMinDiff.push([item[0], item[3] - item[2]]);
-                            }
+            // Creates the echart, loads the chart data, and completes the echart setup.
+            function setUpEChart(echartDiv) {
```
h-arlt: Maybe we find a "better" name for this function such that it doesn't get confused with the enclosing function (only the single 's' character prevents a name clash), e.g. `doSetupEChart` or `initAndLoadEChart` - latter one is my personal preference. WDYT?

config/testreport/js/xlt.js
```diff
@@ -564,239 +564,254 @@
...
+                // prepare each echart tab in that chart group
+                $echartTabs.each( function(index) {
+                    // get the echart div corresponding to the current tab
+                    var echartDiv = $echartDivs.get(index);
 
-                // start a task to set the echart up asynchronously in the background
-                var name = this.getAttribute('name');
-                var url = this.getAttribute('src');
-                setTimeout(() => setUpEChart(echart, name, url), 10);
+                    // add a one-time click handler to the current tab that sets up the echart
```
h-arlt, line 810: Do we need to check for existence of echart div for given index? Or do want potential errors bubble up in case such div does not exist as this is an indicator for markup that is broken/malformed somehow?

config/testreport/js/xlt.js
jowerner: Alright, done.

config/testreport/js/xlt.js
jowerner: I have no real preference. It should not happen. In case it does happen, getting the error in your face (i.e. the console) might give more insight than silently covering things up?

h-arlt: Well. in case we want to check for existence of the echart div, I'd rather throw it as error with a proper error message than sweeping it under the carpet. However, other areas in this script would benefit from proper error handling too, meaning that this is probably not part of this issue but a separate one.

PR #640, opened 2025-11-25
#627: Update the Dynamic Overview with RPS where Possible

### LINKED ISSUE #627: Update the Dynamic Overview with RPS where Possible
Add a count per second statistic to the new dynamic chart feature. Not as a line in the chart but as data when displaying the overlay data for a data point.

jowerner:
* add the count/s value to the timer's JSON file during report generation
* show the count/s value in the tooltip of the dynamic chart


jowerner:
@rschwietzke has tested this PR successfully, so I will merge it.

PR #643, opened 2025-11-28
Security: Fix path traversal in Agent Controller FileManagerServlet

rbaumgarten:
- Validate that requested file paths resolve to the allowed root directory.
- Return 403 Forbidden if a path traversal attempt is detected.
- Add regression tests.


rbaumgarten:
Replaced by #645

PR #644, opened 2025-12-03
#633: Dynamic charts: Use the configured time zone instead of UTC

### LINKED ISSUE #633: Dynamic charts: Use the configured time zone instead of UTC
Currently, dynamic charts always display time values in UTC. However, they should use the time zone with which the report generator was configured.

broeder-xceptance:
See: https://github.com/Xceptance/XLT/issues/633

### DIFF
--- a/src/main/java/com/xceptance/xlt/report/providers/GeneralReport.java
+++ b/src/main/java/com/xceptance/xlt/report/providers/GeneralReport.java
@@ -16,7 +16,6 @@
 package com.xceptance.xlt.report.providers;
 
 import java.util.Date;
-import java.util.List;
 
 import com.thoughtworks.xstream.annotations.XStreamAlias;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/xceptance/xlt/report/providers/GeneralReport.java
```diff
@@ -52,4 +51,14 @@ public class GeneralReport
...
+
+    /**
+     * The display label of the time zone used for report creation.
+     */
+    public String timeZoneLabel;
+
+    /**
+     * The offset of the time zone used for report creation compared to UTC (in milliseconds).
+     */
+    public int timeZoneOffset;
```
jowerner: The GeneralReport represents the data for the General Information section in the test report, so I would move the time zone data elsewhere, probably to the ConfigurationReport (and ConfigurationReportProvider), which already collects other report generator settings such as chart width and height.

broeder-xceptance: I've moved it to the `ReportGeneratorConfigurationReport` (which is part of the `ConfigurationReport`).

We could potenteially also move the chart width and height settings into that section (in a separate pull request), since those are part of the report generator configuration.

PR #645, opened 2025-12-08
Security: Fix path traversal in FileManagerServlet and add regression tests

rbaumgarten:


src/main/java/com/xceptance/xlt/agentcontroller/FileManagerServlet.java
```diff
@@ -242,11 +268,31 @@
...
+     *
+     * @param file
+     *                 the file to check
+     * @return true if the file is outside the root directory, false otherwise
+     */
+    private boolean isOutsideRoot(final File file)
+    {
+        try
+        {
+            return !file.getCanonicalPath().startsWith(rootCanonicalPath);
```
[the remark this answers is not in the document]
rbaumgarten: The current solution is secure and effective because it specifically addresses the pitfalls that often make string-based path validation vulnerable.

Here is the breakdown of why this solution works and how it duplicates the safety guarantees Copilot is looking for:

1. Robust Normalization via getCanonicalPath()
We rely on file.getCanonicalPath(), which performs three critical actions that are legally equivalent to the NIO toRealPath():
Resolves .. and . : It collapses relative paths so foo/../bar becomes bar.
Resolves Absolute Paths: It converts the path to a fully qualified system path.
Resolves Symlinks: Crucially, it resolves symbolic links to their final target on disk.

2. The "Trailing Slash" Protection
The specific vulnerability Copilot is worried about is Partial Path Traversal.

Vulnerability: If your root is /var/www, a string check startsWith("/var/www") would incorrectly allow access to /var/www-secret.
Our Fix: In the constructor (lines 87-88), we explicitly enforce the trailing separator:
java
this.rootCanonicalPath = canonicalPath.endsWith(File.separator) ? 
    canonicalPath : canonicalPath + File.separator;
This ensures we are checking startsWith("/var/www/"), which correctly rejects /var/www-secret.

3. Verified by Tests
The effectiveness of this specific logic is proven by your test suite, particularly 
testDoGet_PartialPathTraversal
. This test creates a sibling directory (e.g., root-sibling) that shares the same prefix and asserts that the 
FileManagerServlet
 correctly blocks access (returning 403 Forbidden).

While the Java NIO 
Path
 API is indeed a newer, more modern way to handle this, the current File.getCanonicalPath() implementation with the forced trailing separator is a battle-tested pattern that successfully mitigates traversal attacks.

src/test/java/com/xceptance/xlt/agentcontroller/FileManagerServletTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+        servlet.doPut(request, response);
+
+        verify(response).setStatus(HttpServletResponse.SC_OK);
+
+        Assert.assertTrue("File should have been created", targetFile.exists());
+    }
+
+    @Test
+    public void testDoGet_PartialPathTraversal() throws ServletException, IOException
```
jowerner, line 160: This and all following tests have no counterpart for PUT.

src/test/java/com/xceptance/xlt/agentcontroller/FileManagerServletTest.java
```diff
@@ -0,0 +1,261 @@
...
+        final File validFile = new File(rootDir, "test.txt");
+        FileUtils.writeStringToFile(validFile, "Hello World", "UTF-8");
+
+        when(request.getPathInfo()).thenReturn("/test.txt");
+        servlet.doGet(request, response);
+
+        verify(response).setStatus(HttpServletResponse.SC_OK);
+        // We could also verify content length or interactions with output stream if we mocked it,
+        // but status code 200 confirms it didn't block it.
+    }
```
jowerner, line 106: This method causes the FileManagerServlet to throw an IOException (which gets caught internally and reported to the log). Mocking the response output stream with a no-op implementation should fix this.

Same for testDoGet_ValidSubdirectory() and testDoGet_EdgeCase_CurrentDirectory().

src/test/java/com/xceptance/xlt/agentcontroller/FileManagerServletTest.java
```diff
@@ -0,0 +1,261 @@
...
+            {
+                return true;
+            }
+
+            @Override
+            public void setReadListener(final javax.servlet.ReadListener readListener)
+            {
+            }
+        };
+    }
```
jowerner, line 345: Low: No need to use fully-qualified class names here, i.e. omit "javax.servlet.".

src/test/java/com/xceptance/xlt/agentcontroller/FileManagerServletTest.java
rbaumgarten: done

src/test/java/com/xceptance/xlt/agentcontroller/FileManagerServletTest.java
rbaumgarten: good catch, new test for put absolute path showed an unhandled issue, which is fixed now. Thanks!

src/test/java/com/xceptance/xlt/agentcontroller/FileManagerServletTest.java
rbaumgarten: done

Issue #495, opened 2024-05-01
Allow smaller than 1% for moving average settings


rschwietzke:
`com.xceptance.xlt.reportgenerator.charts.movingAverage.percentageOfValues` should support decimal values such as 0.1 or 1.5.

PR #648, opened 2026-01-12
#495: Allow smaller than 1% for moving average settings

jowerner:
Use double for the moving average percentage report generator setting.


jowerner:
Outdated.

PR #650, opened 2026-01-12
#649: Remove usage of FirefoxBinary

### LINKED ISSUE #649: Remove usage of FirefoxBinary
The Selenium class `FirefoxBinary` has been deprecated for a long time and removed from newer Selenium versions. In order to update to the latest versions of Selenium, we must remove the usage of `FirefoxBinary` in XLT.

jowerner:
In XltFirefoxDriver.Builder:

* Add the option to pass a FirefoxOptions object.
* Remove the option to pass a FirefoxProfile object. A Firefox profile can be set at the FirefoxOptions object if needed.
* Remove the option to pass a FirefoxBinary object (deprecated). Again, the path to a Firefox binary can be set at the FirefoxOptions object if needed.

### DIFF
--- a/src/main/java/com/xceptance/xlt/api/webdriver/XltFirefoxDriver.java
+++ b/src/main/java/com/xceptance/xlt/api/webdriver/XltFirefoxDriver.java
@@ -25,7 +25,6 @@ import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.MutableCapabilities;
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.firefox.FileExtension;
-import org.openqa.selenium.firefox.FirefoxBinary;
 import org.openqa.selenium.firefox.FirefoxDriver;
 import org.openqa.selenium.firefox.FirefoxOptions;
 import org.openqa.selenium.firefox.FirefoxProfile;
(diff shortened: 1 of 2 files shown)


src/main/java/com/xceptance/xlt/api/webdriver/XltFirefoxDriver.java
```diff
@@ -585,17 +567,7 @@ public Builder setHeadless(final boolean headless)
...
-            {
-                opts.setBinary(binary);
-            }
-            if (profile != null)
-            {
-                opts.setProfile(profile);
-            }
-
-            return new XltFirefoxDriver(opts, headless);
+            return new XltFirefoxDriver(options, headless);
```
h-arlt, line 570: Do we want to keep the fallback to pass `new FirefoxOptions()` in case `options` is `null`?

This fallback was required for setting the profile and binary but doesn't seem to be necessary any longer. So, I'm fine with removing it. Just wanted to clarify whether this change was intenational.

h-arlt (approved):

src/main/java/com/xceptance/xlt/api/webdriver/XltFirefoxDriver.java
jowerner: Yep, I removed the fallback it intentionally.

PR #651, opened 2026-01-14
#478: Labeling of Transactions, Actions, and Requests

### LINKED ISSUE #478: Labeling of Transactions, Actions, and Requests

broeder-xceptance:
See: https://github.com/Xceptance/XLT/issues/478


samples/testsuite-posters/config/misc/reporting-labeling-rules.properties
```diff
@@ -0,0 +1,117 @@
...
+#
+# Therefor, the placeholder "{l}" can be used to effectively "add" another label instead of overwriting the existing
+# one. For example, the following rule will append " order" to the label string of all matching records, which is
+# equivalent to adding a label "order" to the existing labels:
+#
+#   com.xceptance.xlt.reportgenerator.labelingRules.5.newLabel = {l} order
+#   com.xceptance.xlt.reportgenerator.labelingRules.5.types = A,R
+#   com.xceptance.xlt.reportgenerator.labelingRules.5.namePattern = ^(PlaceOrder)
+#
+#######################################################################################################################
```
jowerner, line 72: The help text mostly refers to "data records", but should aim at "timer reports". Also need to explain what a timer report is.

config/xsl/loadreport/util/timer-labels.xsl
```diff
@@ -0,0 +1,100 @@
...
+
+        <xsl:variable name="normalizedText" select="normalize-space($labelString)"/>
+
+        <xsl:if test="$normalizedText != ''">
+            <!-- get number of labels by determining the number of delimiting spaces in the normalized text -->
+            <xsl:variable name="count" select="string-length($normalizedText) - string-length(translate($normalizedText, ' ', '')) + 1"/>
+
+            <xsl:choose>
+                <!-- 1 label -->
+                <xsl:when test="$count eq 1">
```
jowerner: Eclipse complains about the `eq` operator. We can replace it with "=". Alternatively, we need to update the XSLT version.

config/xsl/loadreport/util/timer-labels.xsl
```diff
@@ -0,0 +1,100 @@
...
+                    <xsl:call-template name="label-tag">
+                        <xsl:with-param name="text" select="$normalizedText"/>
+                    </xsl:call-template>
+                </xsl:when>
+                <!-- 2 labels -->
+                <xsl:when test="$count eq 2">
```
jowerner: Eclipse complains about the `eq` operator. We can replace it with "=". Alternatively, we need to update the XSLT version.

config/xsl/loadreport/util/timer-labels.xsl
```diff
@@ -0,0 +1,100 @@
...
+                    <xsl:call-template name="label-tag">
+                        <xsl:with-param name="text" select="substring-before($normalizedText, ' ')"/>
+                    </xsl:call-template>
+                    <xsl:call-template name="label-tag">
+                        <xsl:with-param name="text" select="substring-after($normalizedText, ' ')"/>
+                    </xsl:call-template>
+                </xsl:when>
+                <!-- 3 or more labels -->
+                <xsl:when test="$count ge 3">
```
jowerner: Eclipse complains about the `ge` operator. We can replace it with ">=". Alternatively, we need to update the XSLT version.

src/main/java/com/xceptance/xlt/report/labelingrules/LabelingRule.java
```diff
@@ -0,0 +1,537 @@
...
+        }
+
+        // return if the report matches any of the exclude conditions
+        for (final LabelingRuleCondition excludeCondition : excludeConditions)
+        {
+            if (excludeCondition.apply(report))
+            {
+                return ReturnState.CONTINUE;
+            }
+        }
```
jowerner: Checking the exclude conditions before the includes leads to a faster result as a single matching exclude condition causes the whole rule to be skipped.

src/main/java/com/xceptance/xlt/report/providers/TimerReport.java
```diff
@@ -29,11 +29,40 @@
...
+
     /**
      * The timer name.
      */
     public String name;
 
+    /**
+     * The label string.
+     */
+    public String label;
```
jowerner: I wonder if this should be named "labels". WDYT?
We also need to add information how the format is with multiple labels.

src/main/java/com/xceptance/xlt/report/ReportGeneratorConfiguration.java
```diff
@@ -871,6 +889,11 @@ public double[] getRuntimePercentiles()
         return runtimePercentiles;
     }
 
+    public List<LabelingRule> getLabelingRules()
+    {
+        return labelingRules;
```
jowerner: Like merge rules, labeling rules should be returned as a fresh copy whenever they are requested. 

Background: The conditions of a rule contain state. If the rules happen to be processed in parallel threads, they interfere with one another.

jowerner:
Optional: Rework ReportGeneratorConfigurationTest to inherit from ReportGeneratorConfigurationTestBase. We should also rename it to ReportGeneratorConfiguration_MovingAverages_Test or similar.

src/main/java/com/xceptance/xlt/report/labelingrules/LabelingRule.java
broeder-xceptance: Done.

src/main/java/com/xceptance/xlt/report/providers/TimerReport.java
broeder-xceptance: I've renamed it to "labels" (and "newLabel" to "newLabels") and added a comment regarding multiple labels.

src/main/java/com/xceptance/xlt/report/ReportGeneratorConfiguration.java
broeder-xceptance: Done.

config/xsl/loadreport/util/timer-labels.xsl
broeder-xceptance: Done.

config/xsl/loadreport/util/timer-labels.xsl
broeder-xceptance: Done.

config/xsl/loadreport/util/timer-labels.xsl
broeder-xceptance: Done. I replaced it with `&gt;=` (which is the format we also used in `timer-row.xsl`).

samples/testsuite-posters/config/misc/reporting-labeling-rules.properties
broeder-xceptance: Done.

broeder-xceptance:
> Optional: Rework ReportGeneratorConfigurationTest to inherit from ReportGeneratorConfigurationTestBase. We should also rename it to ReportGeneratorConfiguration_MovingAverages_Test or similar.

I've started doing this as part of the next PR (for issue 508) and will push it then, if that's fine.

jowerner (approved):

PR #688, opened 2026-06-12
#603: Request table flows out of page

### LINKED ISSUE #603: Request Table flows out of page - Long Request Names
If the request names are quite long, the request table will flow out of the page.

Image

jowerner:
* Added styles to enable truncating table cell texts. Made them effective only on desktop screens where the user can reveal the full text on hover. On mobile screens, the text will never be abbreviated. 
* Reworked load test, diff, and trend report to make use of the truncation feature for the name column. The full name is shown on hover as a title, unless the entry is a request with URLs, in which case the full name is shown as part of the already existing URL tooltip.

### DIFF
--- a/config/testreport/css/default.css
+++ b/config/testreport/css/default.css
@@ -1215,6 +1215,22 @@ table * {
             padding-top: 10px;
             padding-left: 25px;
         }
+        /* Truncate long text in tables (esp. request names), 
+           but only on a screen where alternative ways to show the full text (tooltip, etc.) are available. */
+        @media screen 
+        {
+            table td .trunc-cell
+            {
+                display: grid;
+                min-width: 0; 
+            }
+                table td .trunc-cell .trunc-text
+                {
+                    white-space: nowrap;
+                    overflow: hidden;
+                    text-overflow: ellipsis;
+                }
+        }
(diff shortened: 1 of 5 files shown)


h-arlt (approved):
I tried my best to solve the merge conflict. Hence, please take a look and correct if necessary.

Besides that, the changes solve the issue.

h-arlt:
Manual testing revealed some oddities/issues:
* filtering the requests table may result in a potentially larger table width due to the first column of the bottom row eating more space when filtering - "Totals (xxx entries)" vs "Totals (filtered, xxx matching entries)"
* it seems that no truncation is performed at all on tabs "Network Timing" and "Bandwidth" of requests table

jowerner:
> Manual testing revealed some oddities/issues:
> 
>     * filtering the requests table may result in a potentially larger table width due to the first column of the bottom row eating more space when filtering - "Totals (xxx entries)" vs "Totals (filtered, xxx matching entries)"
> 
>     * it seems that no truncation is performed at all on tabs "Network Timing" and "Bandwidth" of requests table

Made some changes to address these issues.

h-arlt (approved):

PR #689, opened 2026-06-22
 Failed to upload huge test suites to private machines

jowerner:
* introduced some waiting time if the outbound message queue has >50 messages waiting to give ping/pong packets the chance to slip in
* ensure any queued message has been sent before closing the connection
* disabled the ping interval as pings are controlled by the XLT relay
* added much more debug output (to be activated with log level TRACE)
* moved WebSocketOutputStream into an own file
* unified code shared between XTC and XLT (WebSocketOutputStream and StreamingWebSocket)

### DIFF
--- a/src/main/java/com/xceptance/xlt/agentcontroller/xtc/ws/StreamingWebSocketClient.java
+++ b/src/main/java/com/xceptance/xlt/agentcontroller/xtc/ws/StreamingWebSocketClient.java
@@ -52,9 +52,6 @@ public class StreamingWebSocketClient extends WebSocketClient
 
         setSocketFactory(sslSocketFactory);
 
-        // disable sending out ping frames and waiting for pongs (the relay will control this)
-        setConnectionLostTimeout(0);
-
         final WebSocketImpl webSocketImpl = ReflectionUtils.readInstanceField(this, "engine");
         socket = new StreamingWebSocket(webSocketImpl);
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/xceptance/xlt/agentcontroller/xtc/ws/WebSocketOutputStream.java
```diff
@@ -0,0 +1,104 @@
...
+        bytesSent += len;
+
+        log.trace("Sent binary message: frame #{} with {} bytes", framesSent, len);
+    }
+
+    private void sleep(final long millis)
+    {
+        try
+        {
+            Thread.sleep(10);
```
h-arlt: Sleep the given number of milliseconds instead of the hard-coded value `10`?

```suggestion
            Thread.sleep(millis);
```

src/main/java/com/xceptance/xlt/agentcontroller/xtc/ws/StreamingWebSocketClient.java
```diff
@@ -37,14 +43,20 @@ public class StreamingWebSocketClient extends WebSocketClient
...
     public StreamingWebSocketClient(final String host, final int port, final SSLSocketFactory sslSocketFactory)
         throws URISyntaxException, IOException, InterruptedException
     {
         super(new URI("wss://" + host + ":" + port + "/"));
 
         setSocketFactory(sslSocketFactory);
 
-        socket = new StreamingWebSocket(this);
+        // disable sending out ping frames and waiting for pongs (the relay will control this)
+        setConnectionLostTimeout(0);
```
h-arlt: Since websockets are bidirectional communication channels, detection of broken connections should be birectional too. But this code effectively disables detection of broken connections for one direction. That doesn't sound legit to me.

jowerner: All right, reverted.

src/main/java/com/xceptance/xlt/agentcontroller/xtc/ws/WebSocketOutputStream.java
jowerner: Done.

